PR #13, opened 2015-02-24
Added LED support to bbb 3.14 config

CaptChrisD:
This will enable the USR LEDS on the Beaglebone black as well as any LEDs that are connected to GPIO pins on capes (provided they have been assigned in the device tree). This config also enables trigger support for the LEDs


fhunleth:
Thanks! I'll try it out. Right now I'm testing out the new version of Buildroot (2015.02) that should be released any day now. I'll try to merge it in with the Buildroot update.

fhunleth:
Merged. See ca5c5f4f0353769ccc84840a4211548a99e63c0a. I ran "make linux-savedefconfig" to merge the configuration options and it was able to simplify the change a little more. The lights blink on my BBB now too, so thanks! Next time that you update your version of nerves, you may want to double check your config files. The new version of Buildroot (2015.02) had a variety of small naming convention changes. I think that the main one that will hit you is that they renamed the linux configurations from linux_xxx.config to linux_xxx.defconfig. You'll see that I renamed the ones that I had in Nerves for the BBB and Raspberry Pi.

PR #18, opened 2015-10-20
Add framebuffer config to get RPi 2 HDMI output working.

fraxu:
RPi 2 didn't give any output from HDMI even when tty1 was set as a console port. Adding similar framebuffer config as in RPi 1 config fixes this.


fhunleth:
Thanks!

PR #19, opened 2015-11-02
Make nerves-env.sh compatible with zsh and others

willpenington:
Update nerves-env.sh to call itself under Bash if used in another shell
to get the correct file location.

Tested in Bash and zsh calling directly, through a symlink, an alias and
PATH.

Equalities also fixed where needed to be compatible with zsh since bash
aliases "==" to "=" and zsh does not.

### DIFF
--- a/nerves-env.sh
+++ b/nerves-env.sh
@@ -3,21 +3,38 @@
 # Source this script to setup your environment to cross-compile
 # and build Erlang apps for this Nerves build.
 
-if [ "$SHELL" != "/bin/bash" ]; then
-    echo ERROR: This script currently only works from bash.
-    exit 1
+# If the script is called with the -get-nerves-root flag it just returns the
+# Nerves SDK directory. This is so that other shells can execute the script
+# without needing to implement the equivalent of $BASH_SOURCE for every shell.
+for arg in $*
+do
+    if [ $arg = "-get-nerves-root" ];
+    then
+        echo $(dirname $(readlink -f $BASH_SOURCE))
+        exit 0
+    fi
+done 
+
+
+if [[ "$SHELL" = "/bin/bash" ]]; then
+    if [ "$0" != "bash" -a "$0" != "-bash" -a "$0" != "/bin/bash" ]; then
+        echo ERROR: This scripted should be sourced from bash:
+        echo
+        echo source $BASH_SOURCE
+        echo
+        return 1
+        exit 1
+    fi
+#elif [[ "$SHELL" = "/bin/zsh" ]]; then
 fi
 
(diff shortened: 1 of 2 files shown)


fhunleth:
Thank you so much for the patch! It looks good, but I haven't gotten a chance to try it out. I'll be able to do that soon and then I'll merge it.

willpenington:
Thanks! Have you got any documentation/cleanup/gruntwork I could
volunteer for? I'd like to get more involved with embedded Linux and
BEAM.

Will


On Thu, Nov 5, 2015, at 01:32 AM, Frank Hunleth wrote:
> Thank you so much for the patch! It looks good, but I haven't gotten
> a chance to try it out. I'll be able to do that soon and then I'll
> merge it.


> — Reply to this email directly or view it on GitHub[1].


>



Links:

  1. https://github.com/nerves-project/nerves-sdk/pull/19#issuecomment-153743736

fhunleth:
Hi Will - Yes, but give me a little more time. Several of us are working on a pretty big change that will make Nerves work better on Mac, and I'm finding that it is introducing small changes everywhere. It made a minor conflict with your script changes for zsh. When I get this sorted out, I'll update and integrate your PR since it is fine. We will certainly need help with creating and documenting starter projects. If you're game to help, I'll ping you after the Mac porting has settled. I assume that you're natively Linux?

willpenington:
I'd be happy to work on those kinds of things so I'll wait to hear from you that the Mac related stuff has settled down. I'm on Arch Linux for my development machine.

nerves-env.sh
```diff
@@ -3,21 +3,38 @@
...
+do
+    if [ $arg = "-get-nerves-root" ];
+    then
+        echo $(dirname $(readlink -f $BASH_SOURCE))
+        exit 0
+    fi
+done 
+
+
+if [[ "$SHELL" = "/bin/bash" ]]; then
```
fhunleth: Line 19 through 29 aren't needed any more, right?

fhunleth:
BTW, I just found out that one of the other Nerves core developers uses zsh on his mac. I merged your change into our working branch, and it got him working natively on OSX. He was thrilled.

willpenington:
That's awesome! I left the code between 19 and 29 so the script fails if it's run instead of sourced in bash (I did that a few times). As you can probably tell from line 28 I was looking at how to add a zsh equivalent too.

fhunleth:
Thanks! I added a comment about implementing this for zsh and other shells. The result is at https://github.com/nerves-project/nerves-sdk/commit/ea0a08074113c5e69d2b5408e57c590f7d9abe0e. It's not in master, but that branch will merged into master as soon as I can verify all of the supported platforms.

PR #20, opened 2015-11-13
Default to on-screen terminal for Elixir on RPi

GregMefford:
I found this to be a more useful default for a newbie.
I think it's more likely that someone would have a monitor attached to the RPi than a serial adapter.


fhunleth:
Thanks for this change as well. I agree that this is the more likely case.

PR #21, opened 2015-11-13
Add dependencies needed on Ubuntu 14.04

GregMefford:
When following these steps on Ubuntu 14.04 Server edition, I also needed these two packages to get through the build process.


fhunleth:
Thanks for the PR for this! I had heard that a couple packages were missing, but I didn't know which ones.

PR #34, opened 2015-12-26
Update nerves_rpi_defconfig to run Erlang shell instead of Elixir shell

mdsebald:
The description of this configuration says this should run the Erlang shell on startup. I modified it to run Erlang shell instead of Elixir shell.  I'm not sure this is entirely correct, but it works for me.


fhunleth:
It looks good from a quick review. I'd merge it, but I know a couple other places that I need to update for consistency too, and I want to double check that projects work. I have a todo to clean up the configs and the documentation from the big non-Linux build refactor work, and your pull request is encouraging me to do that sooner. I'm hanging out with extended family right now, but as soon as I get some time, I'll address this.

Thank you very much for taking the time to point this out!

fhunleth:
I finally got to this. Sorry for the delay. What I did was add a `nerves_rpi_erlang_defconfig` to go along side the `nerves_rpi_elixir_defconfig`. Now there's not a plain `nerves_rpi_defconfig` that's ambiguous. I also updated the docs as you correctly pointed out. Your changes are in `nerves_rpi_erlang_defconfig` as you would expect.

Thanks again for bringing this to my attention. It has been long overdue for me to clean this up.

fhunleth:
I forgot to add that the changes are in 2283176d0a255c69073aaf9340f313c8a3872f83.

PR #36, opened 2016-01-09
Add environment variables to work with erlang.mk

mdsebald:
I know Elixir is nerves' primary language, but I am still working in Erlang, and the Erlang ALE library uses erlang.mk to build.  Erlang.mk, in turn, requires environment variables not currently defined in the nerves environment.  So I added them.  Warning: I did not actually test the changes to "nerves.mk", but it seems a pretty straightforward change, and I added that for completeness.


fhunleth:
Thanks! I squashed the commits and made a minor update to the comments. I'm quite happy that Erlang.mk will work with Nerves now. Even though the core team has been focusing on Elixir, I really appreciate PRs for Erlang (and LFE, etc), since the core codebase really should work well across the Erlang ecosystem and I don't want Elixir-isms creeping into the core if I can help it.

PR #61, opened 2016-03-22
Add Raspberry Pi 3 support.

fairbanksg:
I was able to boot to an iex shell with these changes.

I am not completely happy with the changes to scripts/download-buildroot. I had to use a snapshot version of buildroot that had added rpi3 support, but it appears buildroot snapshots are only available for a short time, so the snapshot version I used will disappear.


fhunleth:
Thanks!

Regarding the Buildroot snapshots, I like to avoid them. There are so many other commits that have been made to Buildroot since the last release that it's a little scary even though Buildroot master is usually pretty stable. My strategy now has been to cherry pick the commits from BR that I want and put them in the patches directory. Could you update your PR with the necessary patches for Pi3 support? I assume that we just need the commits that changed rpi-userland, rpi-kernel and rpi-firmware. The next BR release isn't until June, I think, so we'll be using the patches for a while.

Also, don't worry about the erlang_defconfig. I'd like to limit the _defconfigs going forward to only the elixir one since it's a hassle to maintain all of them and the few that exist can be examples.

The other configs look good.

Finally, could you add the rpi3 to the list of platforms built by Travis?

mobileoverlord:
Looks good @fairbanksg, can you rebase and squash your commits please :smile:  thanks!

fhunleth:
Thanks for the updates! I merged them with a minor change to the patch so that it had a couple comments on what it contained. The commits are 22f92e211c251e74104352009a282fa57b42d163 and 3ea6d2b1616f1bc4a3a0fd0b5d2d436cb073b2b3.

PR #63, opened 2016-04-02
Mix Integration

mobileoverlord:
This is the start of the PR which will allow this repository to be enabled as a Mix Project and checked into Hex.
Things to do
- [x] Add Mix Project and Nerves Configurations
- [x] Relocate `configs/*` into their own repositories. 
- [x] Include the configs as git submodules
- [x] Fix travis to build submodule configurations upon change
- [x] Rename repo `nerves_system_br`


nerves.exs
```diff
@@ -0,0 +1,15 @@
+use Mix.Config
+
+config :nerves_system_br, :nerves_env,
+  type:  :system_build_platform,
+  package_files: [
+    "board",
+    "buildroot-*",
```
fhunleth: The `buildroot-*` directories are automatically generated. Do you really want them here?

mobileoverlord: I don't believe I do, I will change that. The package files I have listed here are the top files and directories to be expanded and recursively included when packaging to send to a remote compiler provider and or to hex. Do you think I need to add / remove anything else

fhunleth:
Manually merged.

PR #74, opened 2016-07-19
scrub-otp-release.sh: Support statically linked executables

arjan:
This lets the `scrub-otp-release.sh` script allow statically linked executable files.


fhunleth:
Good addition. Thanks!

PR #76, opened 2016-07-26
Nerves Package Implementation

mobileoverlord:
Tracked alongside
https://github.com/nerves-project/nerves/pull/27

Work In Progress
PR Includes
* Initial work towards generating manifests
* script/* changes to support initiating from docker to directly produce artifact

Todo

- [x] stabilize dep on `:nerves`


mobileoverlord:
@fhunleth I am ready to make a pass at this. It includes the scripts for generating manifests and the script for generating artifacts using the `make artifact` method. Let me know if you want to break this apart or not.

VERSION
```diff
@@ -1 +1 @@
-0.5.3
+0.8.0-dev
```
fhunleth: Could you break out the version bump into a separate commit?

VERSION
```diff
@@ -0,0 +1 @@
+0.5.3
```
fhunleth: The version 0.5.3 isn't right. I think you fixed it in another commit, but I think that it's worth pulling the VERSION change into a commit all by itself. Also, `nerves.mk` has a version number in it.

external.mk
```diff
@@ -11,21 +11,29 @@ export NERVES_DEFCONFIG_DIR
 # Pull in any configuration-specific packages
 -include $(NERVES_DEFCONFIG_DIR)/external.mk
 
-# Create a system image for use by Bakeware and for creating
-# firmware images without Buildroot
+NERVES_ARTIFACT_NAME ?= $(BR2_NERVES_SYSTEM_NAME)
```
fhunleth: I don't like calling a system an artifact here, since system is more specific as to what it builds.

external.mk
```diff
@@ -11,21 +11,29 @@ export NERVES_DEFCONFIG_DIR
...
-	$(BR2_EXTERNAL)/scripts/mksystem.sh $(BR2_NERVES_SYSTEM_NAME)
+	$(BR2_EXTERNAL)/scripts/mksystem.sh $(NERVES_ARTIFACT_NAME)
 
 # It is common task to copy files to the images directory
 # so that they can be included in a system image. Add this
 # logic here so that a post-createfs script isn't required.
 ifneq ($(call qstrip,$(BR2_NERVES_ADDITIONAL_IMAGE_FILES)),)
 define NERVES_COPY_ADDITIONAL_IMAGE_FILES
-	cp $(call qstrip,$(BR2_NERVES_ADDITIONAL_IMAGE_FILES)) $(BINARIES_DIR)
+  cp $(call qstrip,$(BR2_NERVES_ADDITIONAL_IMAGE_FILES)) $(BINARIES_DIR)
```
fhunleth: The tab character here somehow got converted into 2 spaces.

lib/nerves_system_br/stream.ex
```diff
@@ -0,0 +1,58 @@
+defmodule Nerves.System.BR.Stream do
```
fhunleth: Could I get some comments on stream.ex and also why it's needed?

scripts/mkartifact.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+
+#
+# This script creates a tarball with everything needed to build
+# an application image on another system. It is useful with the
+# `bake` utility so that the parts that require Linux can be
```
fhunleth: bake?

PR #80, opened 2016-09-12
Add environment variables to work with erlang.mk

mdsebald:
Similar to the pull request I made for nerves-env.sh in January '16


fhunleth:
Thanks! I don't get to testing `erlang.mk` projects any more, so I'm really glad that you're sending these changes up.

mdsebald:
Your welcome. Yes, I'm still using erlang_ale which still uses erlang.mk.

PR #81, opened 2016-09-26
Support for Nerves ~> 0.4.0

mobileoverlord:
nerves package config requires version.


fhunleth:
Don't forget to update `nerves-env.sh` and `nerves.mk` for the `VERSION` file change. Believe it or not, they're actually in use. I get PRs for them.

mobileoverlord:
I updated the nerves.mk with the methods suggested.

fhunleth:
Were you going to update `nerves-env.sh`?

mobileoverlord:
I thought nerves-env.sh parsed the version out from the nerves.mk file

https://github.com/nerves-project/nerves_system_br/blob/master/nerves-env.sh#L74

fhunleth:
It does, but it gets set to `$(strip $(shell cat VERSION))` (not evaluated, since it's just text).

mobileoverlord:
Right, that makes sense 😄

mobileoverlord:
I added `scripts/noninteractive-build.sh`
I needed it for the docker provider to perform a non interactive build

Updated the `nerves-env.sh` to read version from the main file

external.mk
```diff
@@ -13,8 +13,9 @@ export NERVES_DEFCONFIG_DIR
 
 # Create a system image for use by Bakeware and for creating
 # firmware images without Buildroot
+NERVES_ARTIFACT_NAME ?= $(BR2_NERVES_SYSTEM_NAME)
```
fhunleth: Can you remove the word ARTIFACT? Pretty, please. That's really the only thing holding me back from merging.

GregMefford: We probably should take out the reference to Bakeware on L14 as well, while you're at it.

scripts/noninteractive-build.sh
```diff
@@ -0,0 +1,40 @@
...
+set -e
+
+NERVES_ARTIFACT_NAME=$1
+NERVES_PKG_PLATFORM=$2
+NERVES_PKG_DEFCONFIG=$3
+NERVES_PKG_OUTPUT=$4
+NERVES_ARTIFACTS_DIR=$5
+
+NERVES_PKG_OUTPUT=/nerves/o/package
+echo "Artifacts Name"
```
GregMefford: I think this should be singular since you're giving the name of just one artifact. Maybe you meant to make it possessive?

fhunleth:
After thinking about this some more, is `scripts/noninteractive-build.sh` really needed? It seems like it just calls `make` a couple times and copies the result. Can't that be done by the program that calls the script? I'm missing why this logic has to be put here.

mobileoverlord:
@fhunleth the script doesn't have to go here, but does need to exist. When building on mac / windows using the Docker provider the host needs a script to call and is then hands off until the exit of the script where it expects an error or an artifact. I put it here because of its nature, its scripting together actions for this repo.

mobileoverlord:
This has now been super de-artifact'ed

fhunleth:
Your latest comment makes me think that a new commit is coming...

PR #84, opened 2016-11-29
update env vars for distillery

mobileoverlord:
instead of referencing an env var with rebar in the key I added a new one which currently points to the same place as the `REBAR_PLT_DIR`. Also added the location to the system_libs instead of trying to resolve it externally.


GregMefford:
@mobileoverlord Are you planning to do anything else on the `env` branch, or can it be cleaned up?

PR #85, opened 2016-12-16
Allow mksystem.sh outside of git

### LINKED ISSUE #82: make system fails when nerves_system_br is not pulled from git
I think we need to _try_ to get this tag
https://github.com/nerves-project/nerves_system_br/blob/master/scripts/mksystem.sh#L51

```
fatal: Not a git repository (or any parent up to mount point /nerves/env/platform)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
make[1]: *** [system] Error 128
/nerves/env/platform/external.mk:18: recipe for target 'system' failed
Makefile:16: recipe for target '_all' failed
make: *** [_all] Error 2
```

jphastings:
Gather the version number for debug from the `VERSION` file instead of the git tag so the build can be completed when being performed from elsewhere.

Fixes #82

### DIFF
--- a/scripts/mksystem.sh
+++ b/scripts/mksystem.sh
@@ -47,9 +47,9 @@ WORK_DIR=$BASE_DIR/tmp-system
 rm -fr $WORK_DIR
 mkdir -p $WORK_DIR/$ARCHIVE_NAME
 
-# Save a tag to the archive in case we need it for debug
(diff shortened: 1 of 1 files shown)


mobileoverlord:
Looks good to me. @fhunleth do you have any thoughts on this?

fhunleth:
Since we moved to the `VERSION` file, this makes sense to me.

PR #87, opened 2017-01-19
Update to Buildroot 2016.11.1

fhunleth:
This brings in a change that affects all Nerves system trees since the
BR2_EXTERNAL variable has been renamed to BR2_EXTERNAL_NERVES_PATH.

System PR's

- [x] https://github.com/nerves-project/nerves_system_bbb/pull/7


fhunleth:
@mobileoverlord The update to Buildroot 2016.11.1 has a big change to BR2_EXTERNAL. Could you review?

mobileoverlord:
From my recollection and a preliminary search across the tools, `BR_EXTERNAL` is only referenced from within this repository. It appears that you have updated all referenced here. We should be 👍

fhunleth:
Here's an example of how it makes everyone update their system images:

https://github.com/nerves-project/nerves_system_rpi/compare/br-bump?expand=1

fhunleth:
One thing that will certainly confuse is `NERVES_SYSTEM` vs. `BR2_EXTERNAL_NERVES_PATH`. They're different things, but I found a lot of instances where the evolution of the variable started making things confusing. I think long ago, NERVES_SYSTEM pointed to nerves_system_br. That no longer works.

mobileoverlord:
yeah I've noticed that in nerves_system_rpi. That was the case at one point. What happened to `BR_EXTERNAL` in the latest BR update?

fhunleth:
The BR people wanted to support multiple BR2_EXTERNAL "layers". To make that happen, they removed the global BR2_EXTERNAL and made everyone name their layer. I named ours "NERVES". This is really cool and we may want to adopt it for the systems, but I didn't want to bite off that refactoring yet until I thought more about it. Another thing that changes due to this is that the "make menuconfig" menus are slightly different now. I had a patch that would put the word "Nerves" in the menu options, but I dropped it with this change. Buildroot now puts "Nerves" in the menus. It's not the same, but I thought that it got the job done well enough to not bother updating the patch.

mobileoverlord:
Sounds like BR2_EXTERNAL "layers" lays the groundwork for composing BR package versions and custom additions from the system level.

fhunleth:
Right.

mobileoverlord:
This all sounds great to me. Lets keep this open and track all system PR's as a todo inside of this one.

fhunleth:
Ok - not quite sure how that's normally done. I've pushed the changes to all of the systems. Here's the PR for the BBB: https://github.com/nerves-project/nerves_system_bbb/pull/7

fhunleth:
Travis succeeds now on all system images. Between @mobileoverlord and I, we've tested BBB and RPi. The updates to Linkit and qemu arm are minor so those shouldn't have broken. Alix, Ag150, Galileo build and pass travis. EV3 builds and passes Travis but has a major kernel update. Since the EV3 takes a long time to verify, I suggest we tag and release. If the EV3 requires fixes, I'll make a new EV3 release. Any reason to delay?

GregMefford:
As far as I'm concerned, it seems like we're ready to shipit. :shipit:
If there are obscure issues on the other platforms, we'll just need to tweak and cut a bug-fix release.

It would be really nice to get a hardware lab established where we can more easily validate against all the "officially-supported" hardware. That should probably be on our roadmap for a 1.0 release.

mobileoverlord:
We can package and ship it. Let's release BR, then tag and build all systems. Thanks @fhunleth 👍

fhunleth:
+1,000,000 to the hardware lab to validate hardware. This process is unbelievably time consuming!

fhunleth:
Ok - I just made a `nerves_system_br` release and pushed it to hex. It's v0.9.0. I'll publish the updated systems tomorrow since it's getting late for me right now.

PR #88, opened 2017-01-27
Pull in all Raspberry Pi related patches from upstream

mobileoverlord:
This includes a kernel bump to 4.4.43 and bumps for rpi-userland and
rpi-firmware. Most importantly, kernels no longer need to be marked that
use device tree, so the kernel marking logic has been removed. This will
break systems, but those systems should upgrade their kernels to avoid
any mismatches between kernel patches and rpi-firmware.


mobileoverlord:
Tested in rpi3 / rpi2 👍

PR #91, opened 2017-04-22
quote when directory may contain special characters or spaces

mobileoverlord:
fixes https://github.com/nerves-project/nerves/issues/114

### DIFF
--- a/scripts/scrub-otp-release.sh
+++ b/scripts/scrub-otp-release.sh
@@ -31,17 +31,17 @@ fi
 # Clean up the Erlang release of all the files that we don't need.
 # The user should create their releases without source code
 # unless they want really big images..
-rm -fr $RELEASE_DIR/bin $RELEASE_DIR/erts-*
+rm -fr "$RELEASE_DIR/bin" "$RELEASE_DIR"/erts-*
 
 # Delete empty directories
-find $RELEASE_DIR -type d -empty -delete
+find "$RELEASE_DIR" -type d -empty -delete
 
 # Delete any temp files, release tarballs, etc from the base release directory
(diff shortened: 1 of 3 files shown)


mobileoverlord:
While doing this, I noticed that the scripts are being referenced from the artifact. Therefore, this change would need to be propagated across all systems to take effect. Do we need to continue to copy these scripts to the system artifact or should we reference them in `nerves_system_br`?

scripts/scrub-otp-release.sh
```diff
@@ -31,17 +31,17 @@ fi
 # Clean up the Erlang release of all the files that we don't need.
 # The user should create their releases without source code
 # unless they want really big images..
-rm -fr $RELEASE_DIR/bin $RELEASE_DIR/erts-*
+rm -fr "$RELEASE_DIR/bin" "$RELEASE_DIR/erts-*"
```
fhunleth: This quoting won't work since the `*` is in the quotes. Here's an example:
```
$ touch foo-1 foo-2
$ ls
foo-1  foo-2
$ echo $PWD
/home/fhunleth/tmp
$ rm "$PWD/foo*"
rm: cannot remove '/home/fhunleth/tmp/foo*': No such file or directory
```
The fix is to change this to `rm -fr "$RELEASE_DIR/bin" "$RELEASE_DIR"/erts-*`

mobileoverlord:
Tested this with a full build through to mix firmware in a directory `/path/to +/my_app` Looks good 👍

fhunleth:
Thanks - I merged it. I'll make a `nerves_system_br` release so that it can start propagating to nerves system images.

PR #97, opened 2017-07-17
replace permissions case implementation with awk

radu:
The case implementation broke when adding the `sudo' buildroot package, when a folder had permission is drwx--x--x. 
This awk implementation works.

### DIFF
--- a/scripts/merge-squashfs
+++ b/scripts/merge-squashfs
@@ -23,33 +23,11 @@ else
     STAT=stat
 fi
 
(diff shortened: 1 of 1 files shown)


fhunleth:
Much better. Thanks for replacing that code.

fhunleth:
See comment in #99. I had to revert this.

radu:
apologies for regression, submitted new change that handles
setuid/gid/sticky

PR #101, opened 2017-07-20
Add python to dependencies

orestis:
Ubuntu Server 16.04 doesn't come with python installed, so needs to be made explicit. I am not sure about CentOS/RHEL, but doesn't hurt to add it.


fhunleth:
Thanks!

PR #102, opened 2017-07-21
match all possible permissions, no more chmod huh

radu:
done in octal as requested, because there's guaranteed no rollovers we could have kept decimal representation.


fhunleth:
Thanks! I merged it into master. It will be in the v0.13.3 release.

PR #107, opened 2017-08-14
move nerves package config from nerves.exs to mix.exs

mobileoverlord:
Updated for use with https://github.com/nerves-project/nerves/pull/176


mobileoverlord:
merged in master

PR #115, opened 2017-10-24
Skeleton missing resolv.conf symlink

mobileoverlord:
add symlink for /etc/resolv.conf -> /tmp/resolv.conf


fhunleth:
Whoa, that's not good. Thanks for fixing!

PR #118, opened 2017-11-09
Allow for custom patches to be added.

ConnorRigby:


ConnorRigby:
This has been solved in a different method and i don't believe we will need this. Feel free to close.

PR #122, opened 2017-12-07
Docker

mobileoverlord:
Add docker files for automation tools and nerves tooling

### DIFF
--- a/support/docker/nerves/Dockerfile
+++ b/support/docker/nerves/Dockerfile
@@ -1,0 +1,39 @@
+FROM nervesproject/nerves_system_br:latest
+LABEL maintainer="Nerves Project developers <nerves@nerves-project.org>" \
+      vendor="NervesProject" \
+description="Container with everything needed to build Nerves projects"
+
+# Setup environment
+ENV DEBIAN_FRONTEND noninteractive
+ENV LANG=C.UTF-8 \
+  TERM=xterm \
+  ERLANG_OTP_VERSION=20.0 \
+  FWUP_VERSION=0.15.4
+# The container has no package lists, so need to update first
+RUN set -xe \
+  && ERLANG_PKG='erlang-solutions_1.0_all.deb' \
+  && ERLANG_URL="https://packages.erlang-solutions.com/${ERLANG_PKG}" \
+  && apt-get update \
+  && apt-get -y install \
+    curl \
+    git \
+    openssh-client \
+    build-essential \
+    libssl-dev \
+    bc \
+    squashfs-tools \
+    gnupg \
+  && curl -o "/tmp/${ERLANG_PKG}" ${ERLANG_URL} \
+  && dpkg -i "/tmp/${ERLANG_PKG}" \
+  && apt-get update \
+  && apt-get -y install \
+    "esl-erlang=1:${ERLANG_OTP_VERSION}" \
+  && rm -rf /var/lib/apt/lists/* \
+  && wget https://github.com/fhunleth/fwup/releases/download/v${FWUP_VERSION}/fwup_${FWUP_VERSION}_amd64.deb \
+  && dpkg -i fwup_${FWUP_VERSION}_amd64.deb \
+  && rm -f *.tar.gz \
+  && rm -f fwup_${FWUP_VERSION}_amd64.deb \
+  && mkdir -p /root/.ssh \
+  && ssh-keyscan -t rsa github.com > /root/.ssh/known_hosts \
+  && chmod 700 /root/.ssh \
+  && chmod 600 /root/.ssh/known_hosts
--- a/support/docker/nerves/Dockerfile
+++ b/support/docker/nerves/Dockerfile
@@ -5,14 +5,14 @@ description="Container with everything needed to build Nerves projects"
 
 # Setup environment
 ENV DEBIAN_FRONTEND noninteractive
-ENV LANG=C.UTF-8 \
-  TERM=xterm \
-  ERLANG_OTP_VERSION=20.0 \
-  FWUP_VERSION=0.15.4
+ENV LANG=C.UTF-8 
+ENV TERM=xterm 
+ENV ERLANG_OTP_VERSION=20.0
+ENV FWUP_VERSION=0.18.0
+ENV ERLANG_PKG='erlang-solutions_1.0_all.deb'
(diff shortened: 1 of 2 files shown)


support/docker/nerves_system_br/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
+    locale-gen
+ENV LANG en_US.UTF-8  
+ENV LANGUAGE en_US:en  
+ENV LC_ALL en_US.UTF-8
+
+ENV DEBIAN_FRONTEND noninteractive
+# The container has no package lists, so need to update first
+RUN apt-get update -y -qq
+RUN apt-get install -y \
```
GregMefford: We should combine all these `RUN` lines into one with `&&` so it only creates one Docker filesystem layer with only the packages installed and all the `apt` caches cleaned-up.

support/docker/nerves/Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM nervesproject/nerves_system_br:latest
+LABEL maintainer="Nerves Project developers <nerves@nerves-project.org>" \
+      vendor="NervesProject" \
+description="Container with everything needed to build Nerves projects"
+
+# Setup environment
+ENV DEBIAN_FRONTEND noninteractive
+ENV LANG=C.UTF-8 \
+  TERM=xterm \
```
GregMefford: Maybe these should be separate `ENV` lines?

support/docker/nerves/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+  ERLANG_OTP_VERSION=20.0 \
+  FWUP_VERSION=0.15.4
+# The container has no package lists, so need to update first
+RUN set -xe \
+  && ERLANG_PKG='erlang-solutions_1.0_all.deb' \
+  && ERLANG_URL="https://packages.erlang-solutions.com/${ERLANG_PKG}" \
```
GregMefford: Is there a reason to not have these be on `ENV` lines?

support/docker/nerves_system_br/Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM buildroot/base
+LABEL maintainer="Nerves Project developers <nerves@nerves-project.org>" \
+      vendor="NervesProject" \
+description="Container with everything needed to build Nerves Buildroot Systems"
+
+# Setup environment
+# Set the locale
+RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
+    locale-gen
+ENV LANG en_US.UTF-8
```
GregMefford: Se this one to `C.UTF-8` and remove the `LANGUAGE` and `LC_ALL` lines.

rameez03online:
Hai Linux Gurus GregMefford and mobileoverlord, I reached this thread based on a messy problem I was having with a group of related errors 1)can not install locale invalid argument
2)warning: setlocale: LC_ALL: cannot change locale 
I think you people found out the problem and its fix. I can see some posts above but confused that what should I do to fix it. Please anyone tell me how average users like me can use these fixes. I am running behind a problem for the last one week.Thanking you in advance.

mobileoverlord:
You can fix the issue by running `docket pull nervesproject/nerves_system_br` this will pull the latest docker image

rameez03online:
Should I need to install any packagr for this? Eg: "docket"

mobileoverlord:
Sorry, typo on my phone. `docker pull nervesproject/nerves_system_br`

support/docker/nerves_system_br/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+LABEL maintainer="Nerves Project developers <nerves@nerves-project.org>" \
+      vendor="NervesProject" \
+description="Container with everything needed to build Nerves Buildroot Systems"
+
+# Setup environment
+# Set the locale
+RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8 C.UTF-8/' /etc/locale.gen && \
+    locale-gen
+ENV LANG C.UTF-8
+ENV LC_ALL C.UTF-8
```
GregMefford: I think we actually don't need the `sed` line or the `LC_ALL` line, but it doesn't really hurt anything either.

GregMefford: Oh wait maybe we do need that `sed` line if it's not already uncommented by default.
The `LC_ALL` is just used to override `LANG`, though, as far as I know.

rameez03online: Hi mobileoverlord,
I ran "docker pull nervesproject/nerves_system_br"
and reported 
Digest: sha256:66e9bc0a826236d0e15f7ef0b25b342388c7d94cab4d71ae1ea30fde664de836
Status: Downloaded newer image for nervesproject/nerves_system_br:latest
after that I ran "locale -a" and found the same error
[rameez@nebuchedneiser ~]$ locale -a
locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_COLLATE to default locale: No such file or directory
C
POSIX
Is it ok?

rameez03online:
Hi mobileoverlord,
I ran "docker pull nervesproject/nerves_system_br"
and reported
Digest: sha256:66e9bc0a826236d0e15f7ef0b25b342388c7d94cab4d71ae1ea30fde664de836
Status: Downloaded newer image for nervesproject/nerves_system_br:latest
after that I ran "locale -a" and found the same error
[rameez@nebuchedneiser ~]$ locale -a
locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_COLLATE to default locale: No such file or directory
C
POSIX
Is it ok?

mobileoverlord:
@rameez03online Are you using Docker with Nerves or are you just on a linux distro using `guix`? It seems like you might be just using the latter, in which case, this thread would not help you. If you are using Nerves, but on a different linux distro, you would have to find information on setting the locale for that distro.

rameez03online:
Yes you are right I am using arch linux and I just installed guix package manager. Sorry to disturb you people.. All the best for your ongoing works.

PR #123, opened 2017-12-11
Buildroot: bump to 2017.11

fhunleth:
This makes it possible to remove all of the QT5 Webengine patches since
they've been upstreamed. The 0012-erlang.. patch fixes an issue with
building host-erlang on systems without libz-dev installed. This patch
is in Buildroot master.

The main change that affects systems is the u-boot bump. I've forced the u-boot version on affected systems for the near-term.


fhunleth:
@mobileoverlord Thanks! I'll merge then!

PR #126, opened 2017-12-14
Resave the Busybox options

fhunleth:
Busybox 1.27.2 added a few more options. This synchronizes the version
that we use so that enabling/disabling single commands doesn't result in
a lot of deltas.


board/nerves-common/busybox-1.22.config
```diff
@@ -330,11 +345,6 @@ CONFIG_FEATURE_VERBOSE=y
 #
 CONFIG_FEATURE_PRESERVE_HARDLINKS=y
 
-#
-# Common options for ls, more and telnet
-#
-CONFIG_FEATURE_AUTOWIDTH=y
```
GregMefford: Was this intentionally removed?
Maybe this matches with the new line below?

```diff
+CONFIG_FEATURE_CMDLINE_MODULE_OPTIONS=y
```

board/nerves-common/busybox-1.22.config
```diff
@@ -1024,24 +1048,23 @@ CONFIG_BASH_IS_NONE=y
...
-CONFIG_ASH_BUILTIN_ECHO=y
-# CONFIG_ASH_BUILTIN_PRINTF is not set
-# CONFIG_ASH_BUILTIN_TEST is not set
+# CONFIG_ASH_RANDOM_SUPPORT is not set
+# CONFIG_ASH_EXPAND_PRMT is not set
+# CONFIG_ASH_IDLE_TIMEOUT is not set
+# CONFIG_ASH_MAIL is not set
+CONFIG_ASH_ECHO=y
+CONFIG_ASH_PRINTF=y
+CONFIG_ASH_TEST=y
```
GregMefford: Looks like these two were not enabled before. Just making sure.

board/nerves-common/busybox-1.22.config
fhunleth: It looks like they removed `CONFIG_FEATURE_AUTOWIDTH` and added a new one for each command. Autowidth is disabled for all of the commands now. That's not intentional. However, I'm not sure that I even used autowidth before. I remember `ls` being pretty much unusable in the runtime shell due to it dumping all of the files in one column and sorted from shortest name to longest. `more` and `telnet` weren't enabled.

board/nerves-common/busybox-1.22.config
fhunleth: I think these must be new options. The `config-pin` script on the BBB uses them.

PR #128, opened 2018-01-10
Support rpi-firmware kernel selection

fhunleth:
This makes it possible to support Raspberry Pi systems that use Linux kernel 4.4 or 4.9 rather than making them go lock-step with `nerves_system_br`. 4.4 is the default to avoid breaking existing systems that don't specify the kernel.


fhunleth:
The only thing that I'm not sure about is using the same `rpi-userland` for both kernels. I hadn't been doing that in the past. What I did was use the `rpi-userland` that came out at the same time as the `rpi-firmware`. I don't think that this was necessary. I was concerned about there being APIs in the firmware blobs that userland would invoke, but I've gotten less convinced that there's a problem. I'll be testing that more after the upcoming system release.

PR #135, opened 2018-03-14
Erlang: bump to 20.3

fhunleth:


CaptChrisD:
FYI there is a bug in 20.3 around ssl. There is already a patch discussed and an update will probably be released very shortly to resolve. There has been a lot of discussion on the Erlang-questions mailing list

fhunleth:
@CaptChrisD Thanks for letting me know. I'll hold off on making a `nerves_system_br` release.

PR #144, opened 2018-04-11
buildroot: bump to 2018.02.1

fhunleth:
This is patch release to 2018.02. This lets us drop two patches that
previously were needed for building newer Linux kernels. Here's the list
of changes from the announcement:

Buildroot 2018.02.1 is a bugfix release on the current long term
release, fixing a number of important / security related issues
discovered since the 2018.02 release.

- Infrastructure: Speed up instrumentation for package size by
 comparing timestamps instead of md5 checksums, as doing checksums
 could lead to up to 25% longer build time for big configurations.

- Add a better fix for build issues for autotools based packages
 checking for C++ support on toolchains without C++ support.

- Build host-tar if tar is older than 1.27 on the build machine to work
 around reproducibility issues with git archives containing long paths.

- check-unique-files: Fix for filenames not representable in the users'
 locale.

- Check-bin-arch: Add support for per-package ignore paths
 (<pkg>_BIN_ARCH_EXCLUDE).

- linux: Add _NEEDS_HOST_OPENSSL / LIBELF options to ensure these
 dependencies are available for Linux configurations needing these
 dependencies, similar to how it is done for U-Boot.

- linux: Support custom (not in tree) device tree files in addition to
 in-tree ones.

- Removed unmaintained defconfigs: Creator Ci40, Firefly rk3288,
 Freescale mpc8315erdb / p1010rdb_pa, Riotboard, Teliv EVK Pro3.

- Security fixes for apache, imagemagick, irssi, libcurl, libopenssl,
 libpjsip, libvorbis, memcached, nodejs, ntp, patch, python-webpy,
 samba4, tremor, wireshark, xerces

- Fixes for download/compilation/runtime issues in busybox, c-ares,
 cmake, docker-engine, enlightenment, eudev, fbgrab, freescale-imx,
 gtest, intel-microcode, jimtcl, kmod, kodi, ktap, libcoap, libfcgi,
 libhttpparser, libminiupnpc, libss7, libssh2, libupnp, libupnpp,
 libuv, motion, musl, nodejs, ntp, openblas, opencv3, openocd,
 opus-tools, php, pkgconf, postgresql, powerpc-utils,
 python-jsonschema, python-webpy, qemu, qt5bsae, qt5script, qt5webkit,
 rapidjson, rauc, sam-ba, sngrep, systemd, uboot-tools, uclibc-ng-test,
 upmpdcli, xterm, zstd.

- New packages: docker-proxy, python-functools32


fhunleth:
My set of Nerves systems all built with this update. This isn't too surprising since not much changed. It looks like I do get some false warnings from Buildroot on its file installation check for the openssl package. I suspect that I missed seeing them before.

ConnorRigby:
Funnily enough I built a custom system with this already. Worked fine :+1:

PR #156, opened 2018-05-10
boardid: bump to v1.0.0

fhunleth:
This brings in support for reading serial numbers from U-boot
environment variables and letting user specify multiple ways of
generating a unique ID.


ConnorRigby:
I tried this on both BeagleBoneGreen wireless and BeagleBonePocket, and `boardid` Seems to work :+1:

PR #158, opened 2018-05-23
qt5 bump to 5.10.1

mobileoverlord:
built and booted using `kiosk_sytsem_rpi3`


fhunleth:
I assume that the non-Qt5 patches are required now to build? Also, could you rerun the patch generation with `git format-patch -N 2018.02.2`? There's some minor cruft in the patch headers that the `-N` will remove.

mobileoverlord:
The other patches are also required now to build Qt. I force pushed the updated patches.

PR #167, opened 2018-06-10
qt5webengine: fix broken enable condition

### LINKED ISSUE #166: Can't use qtwebengine with QT 5.10
After upgrade to `nerves_system_br` `1.2` there is no way to use `qtwebengine` with `QT` `5.10`. You have to downgrade to `5.6 (TLS)`. `qtwebengine` does not appear in `menuconfig`.

Found something like this: `Depends on: ... && (BR2_TOOLCHAIN_HAS_GCC_BUG_85862 [=n] || BR2_PACKAGE_QT5_VERSION_5_6 [=n]) && ...` 

Tried to switch `gcc` to `7.X` but with no luck.

fhunleth:
This should fix #166. I have not tested it, though.


mobileoverlord:
Tested on kiosk_system_rpi3. Booted to browser. Also verified that the entry shows up in `make menuconfig`

PR #178, opened 2018-07-03
Adds an additional solution to the Erlang OTP version error.

pdgonzalez872:
Closes https://github.com/nerves-project/nerves_system_br/issues/177


fhunleth:
Much better. Thanks!

PR #182, opened 2018-07-06
nbtty: bump to v0.4.0

fhunleth:
This pulls in a fix for process exits not being detected and support for
retrying tty files so that progress can be made on configfs/USB gadget
work.


mobileoverlord:
Oh neat. Tty polling is in. That’s cool.

PR #185, opened 2018-07-06
Pull in qt5webengine parallel compile fix

fhunleth:
I have not tested this.


mobileoverlord:
I'll test this later today in the kiosk systems and report back.

PR #186, opened 2018-07-06
Fix issue with calling readelf and give better error

fhunleth:
This fixes the scrubber failing when releases contain shell scripts.


mobileoverlord:
Hooray for better error messages 👍

PR #190, opened 2018-07-13
Enable `ln` in Busybox

GregMefford:
This is the first step in enabling `configfs` and ultimately solving the following:
* https://github.com/nerves-project/nerves_system_rpi0/issues/29
* https://github.com/nerves-project/nerves_system_rpi0/issues/45
* https://github.com/nerves-project/nerves_system_rpi0/issues/46
* https://github.com/nerves-project/nerves_runtime/pull/58

We had high hopes that OTP21 would allow us to directly create symlinks in `configfs` paths from Elixir (using `File.ln_s!`), but it does not work. Writing the other files under `configfs` (using `File.write!`) does work in OTP 21, though.


GregMefford:
Actually, never mind! I was just mis-using `File.ln_s!` in that you have to tell it the file name you want the symlink to have in the second argument (which is different than `ln -s` on the command-line, which infers the same file name as the source).

Hooray for OTP 21! 🎉

PR #197, opened 2018-07-31
fwup: bump to v1.2.4

fhunleth:


fhunleth:
FYI - the fix in 1.2.4 doesn't affect any Nerves projects to my knowledge, but the bug was painful so I want to make the fix available everywhere.

PR #198, opened 2018-07-31
Support Qt/QMake projects

fhunleth:
This saves the mkspecs directory to staging so that it gets distributed
with systems and then sets the QMAKESPECS environment variable for
builds. `qmake` will look at the variable to know how to build Qt-based
projects.


mobileoverlord:
This is an exciting one :)

PR #207, opened 2018-08-08
Create and switch user in docker entrypoint if one was set

mobileoverlord:
When running the docker image, we set the user and group ids to those of the user from the host. This fixes permissions issues that arise from writing files back to the host from within the container. Unfortunately since this user and group do not actually exist, it can cause some bad behavior with packages in buildroot. In the example case, `host-ncurses`. This will set the entrypoint and create and switch users if the values were passed through in the `--env`


mobileoverlord:
Fixes https://github.com/nerves-project/nerves_system_br/issues/206

fhunleth:
It looks like Buildroot's behavior might change in the future so that we no longer need this:

http://lists.busybox.net/pipermail/buildroot/2018-August/227214.html

PR #214, opened 2018-08-14
erlang: continue to support OTP 20 systems

fhunleth:
This adds support for OTP 20 back so that users who cannot upgrade to
OTP 21 can still follow along with the latest patches and security
updates.

The default remains OTP 21. OTP 20 must be explicitly chosen.


ConnorRigby:
Wow that was really fast. Good job 💯

fhunleth:
Thanks - I probably could do more testing, but it's low risk and I'm working down my pre-ElixirConf bug fix list.

PR #221, opened 2018-08-20
Docker image updates

mobileoverlord:
Trying to accomplish the following
1. Fix an issue with docker running on linux hosts where it would raise an exception that the group id for 1000 does not exist. This is an issue in the docker-entrypoint script. The PR adds the group before trying to add the user to it.
2. Improve docker image layer caching. Since fwup changes more frequently then the reset of the host packages, I am splitting it out in an attempt to see if it can reuse the other layers. I am hoping that this shrinks an 800mb download on every bump to roughly the size of fwup. 
3. Add `libmnl-dev` for building nerves project apps using this image.

### DIFF
--- a/support/docker/nerves_system_br/Dockerfile
+++ b/support/docker/nerves_system_br/Dockerfile
@@ -68,16 +68,13 @@ RUN dpkg --add-architecture i386 \
     bc \
     squashfs-tools \
     gnupg \
+    libmnl-dev \
   && curl -o "/tmp/${ERLANG_PKG}" ${ERLANG_URL} \
   && dpkg -i "/tmp/${ERLANG_PKG}" \
   && apt-get update \
   && apt-get -y install \
     "esl-erlang=1:${ERLANG_OTP_VERSION}" \
   && rm -rf /var/lib/apt/lists/* \
-  && wget https://github.com/fhunleth/fwup/releases/download/v${FWUP_VERSION}/fwup_${FWUP_VERSION}_amd64.deb \
-  && dpkg -i fwup_${FWUP_VERSION}_amd64.deb \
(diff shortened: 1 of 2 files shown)


michaelkschmidt:
Each "run" statement creates a unique layer that depends on the layers above it.  If you have something small like fwup that changes alot, move it to the the bottom of the file.  That way it only has to download the "fwup run" layer each time it changes

mobileoverlord:
@michaelkschmidt Ah thanks, I am not sure then if this is going to work. Even though moving it to the bottom would make it smaller. When building the new image with a fwup update it will run the layer above which has a call to `apt-get`. I believe that this causes enough of a ripple each time to mess up the layer cache. Is this true, and if so, is there a way around it?

michaelkschmidt:
Docker actually ignores the fact that `apt-get` might change from run to run.  Unless you explicitly do something to invalidate that layer's cache, it will assume its the same.

For our docker stuff, we always add to the bottom of the file, and then periodically move things higher.  This allows quick updates until we are forced to re-push the world

support/docker/nerves_system_br/Dockerfile
```diff
@@ -50,7 +63,6 @@ RUN dpkg --add-architecture i386 \
     bzip2 \
     bison \
     flex \
-    wget \
```
fhunleth: Removing `wget` feels like it might break some scripts.

mobileoverlord: It was moved higher in the script.

mobileoverlord:
Let me fix this up before we release a new system_br

mobileoverlord:
Updated, ready for review

PR #230, opened 2018-08-27
Set the group and user even if it exists

mobileoverlord:
use `-o` to fix issues with setting groups that already exist `groupadd: GID '20' already exists`


fhunleth:
:man_shrugging:

PR #233, opened 2018-09-06
Use single sed call with multiple -e's

jmbutter:
Piping multiple sed calls together can result in extreme slowness on some systems.  However, the same result can be accomplished more efficiently by using a single call to sed with multiple scripts passed in (see https://www.gnu.org/software/sed/manual/sed.html#Multiple-commands-syntax)

This change reduced `mix firmware` runtime from 30+ minutes to < 2 minutes on my system (MacOS High Sierra 10.13.6).

### DIFF
--- a/scripts/merge-squashfs
+++ b/scripts/merge-squashfs
@@ -54,7 +54,7 @@ owner_to_uid_gid() {
 
 get_path() {
     # trim the squashfs-root and escape any spaces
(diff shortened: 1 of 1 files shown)


jmbutter:
This works on MacOS, but should probably be tested on Linux / other supported OS's.

fhunleth:
Thanks!!!! It looks like it works on Linux.

axelson:
It's neat that you can pass multiple `-e` arguments to sed.

PR #234, opened 2018-09-11
Buildroot: bump to 2018.08

fhunleth:
See http://buildroot-busybox.2317881.n4.nabble.com/Buildroot-2018-08-released-td203016.html.

A host-dtc change causes an error when building Nerves systems with old
versions of U-Boot. Here's important text from the above release note:

  Beware of the following known issues:

  - Glibc 2.28 on the build host breaks compilation of a number of host
    packages. 2018.08 contains fixes for some of these packages, but not
    all. Consider building on hosts (or in containers) using older Glibc
    versions.

  - host-dtc 1.4.7 breaks compilation of older U-Boot and Linux kernel
    configurations using FDT/DTC. Consider updating the Linux kernel to >=
    4.17 and U-Boot to >= 2018.07 or backporting commit 9130ba8846
    (scripts/dtc: Update to upstream version v1.4.6-9-gaadd0b65c987) for
    the Linux kernel / commit db405d1980 for U-Boot.
    Alternatively revert commit 7b929ddcf0 (dtc: bump version to 1.4.7)
    and ensure your build host does not have the libfdt development
    headers installed.


fhunleth:
This change currently breaks `nerves_system_bbb`. The other systems that I have locally build. The Linux 4.14 update to `nerves_system_bbb` works.

mobileoverlord:
I guess we will have to land the 4.14 BBB

PR #237, opened 2018-09-11
erlang: refactor patches and bump to 21.0.8

fhunleth:
The refactoring of the patches breaks the erlang changes into finer
grain commits to make it easier to maintain. Bumping the version was
getting painful in terms of rebasing conflicts. This makes it easier and
also makes it more possible to push our Erlang patches upstream.

### DIFF
--- a/patches/buildroot/0010-erlang-switch-to-Github.patch
+++ b/patches/buildroot/0010-erlang-switch-to-Github.patch
@@ -1,0 +1,29 @@
+From 0fd42ef0d860a33eac4ff3fe6b9e8ce76bb6e836 Mon Sep 17 00:00:00 2001
+From: Frank Hunleth <fhunleth@troodon-software.com>
+Date: Tue, 11 Sep 2018 11:54:25 -0400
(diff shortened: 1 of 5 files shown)


patches/buildroot/0010-erlang-switch-to-Github.patch
```diff
@@ -0,0 +1,29 @@
...
+index 6dd5e85e76..5a39d45a81 100644
+--- a/package/erlang/erlang.mk
++++ b/package/erlang/erlang.mk
+@@ -6,8 +6,8 @@
+ 
+ # See note below when updating Erlang
+ ERLANG_VERSION = 21.0
+-ERLANG_SITE = http://www.erlang.org/download
+-ERLANG_SOURCE = otp_src_$(ERLANG_VERSION).tar.gz
++ERLANG_SITE = https://github.com/erlang/otp/archive
```
ConnorRigby: Do you happen to have a build before and after this one? i remember erlang.org having really slow download speeds.

PR #246, opened 2018-09-21
Revert to dtc 1.4.4

fhunleth:
The steps to make Linux kernels before 4.17 work with dtc 1.4.7 was too
frustrating, so I gave up. I did try other versions of dtc. dtc 1.4.5
crashes and 1.4.6 fails similar in a similar manner to 1.4.7.


fhunleth:
sigh. This breaks host-uboot-tools.

PR #265, opened 2018-11-26
erlang: bump to 21.1.3

fhunleth:
This is the only change:

```
 OTP-15430    Application(s): erts
               Related Id(s): ERIERL-237

               Added an optional ./configure flag to compile the
               emulator with spectre mitigation:
               --with-spectre-mitigation

               Note that this requires a recent version of GCC with
               support for spectre mitigation and the
               --mindirect-branch=thunk flag, such as 8.1.
```


fhunleth:
Note that the Nerves toolchains currently use gcc 7.x, so based on the changelog, this shouldn't fix anything for us yet.

PR #276, opened 2018-12-12
openssl: bump to 1.1.0h

fhunleth:
This pulls in a pending patch in Buildroot to update the version of
OpenSSL from 1.0.2 to 1.1.0h. This fixes what appears to be issues with
Erlang using OpenSSL engines. It also enables Erlang crypto features
like additional algorithms.

OpenSSL 1.0.2 is also feeling dated now. The OpenSSL project has
indicated that they'll stop supporting it after 2019.


fhunleth:
This could break 3rd party libraries. One would hope that everyone has upgraded to OpenSSL 1.1.0 by now, but I recently came across with one library that provided an SSL engine that had not.

PR #294, opened 2019-01-13
boardid: bump to v1.5.3

fhunleth:


fhunleth:
This fixes an issue with overriding config files using the commandline.

PR #300, opened 2019-02-06
Disable host-erlang since unneeded

fhunleth:
This saves minutes of build time and is unneeded for Nerves since Erlang
source code is compiled outside of Buildroot.


fhunleth:
It's historical. It hasn't been needed for at least a year.

PR #305, opened 2019-02-19
Bump fwup version for Docker

fhunleth:


fhunleth:
I forgot to bump the `fwup` version for the Docker image. This means that builds on CircleCI will fail if anyone uses the new `expand` option in their `fwup.conf` files. Do we need to make a new `nerves_system_br` release to push this out?

mobileoverlord:
yeah, they would have to have a tagged version of nerves_system_br to reference.

PR #318, opened 2019-04-04
erlang: bump to 21.3.3

fhunleth:
See http://erlang.org/download/OTP-21.3.3.README


fhunleth:
I'm not running regularly with 21.3.x builds. I'll probably hold off merging until I try it out for a bit.

fhunleth:
I’ve been running this without issues. I think it’s ok.

PR #322, opened 2019-05-31
erlang: bump to 22.0.1

fhunleth:
This also reorganizes the patch files to make this easier to maintain.
Users can downgrade their version of Erlang to an OTP 21 or OTP 20
release via menuconfig.


jjcarstens:
👍 - Just a random thought: Is it worth updating docs somewhere that one can now change Erlang version with `menuconfig`?

fhunleth:
Maybe. I've been quiet about this feature since I don't maintain the old versions. It only exists since one person needed it a year ago. Where would you expect to see it? (Asking honestly, since I'm so close to Buildroot that this location is the ideal location and I wouldn't think to look anywhere else)

mattneel:
I use the older version of OTP due to being too lazy to update a dep using a port that breaks with the newer changes. It took me all of five minutes to find it in menuconfig, though I'm pretty familiar with buildroot.

I found the feature via the commit log, so maybe a small aside somewhere in the docs would be warranted for others in the same boat. If you point me to where a good place to stick it would be, I could find some time after I finish moving today to do it myself so you guys can focus elsewhere.

fhunleth:
Thanks! Maybe at the top of this doc? https://github.com/nerves-project/nerves/blob/master/docs/Customizing%20Systems.md

jjcarstens:
☝️ Thats exactly where I was thinking. Mainly because as a very inexperienced Buildroot user myself, that's where my mind initially gravitated to if I were to ask "How can I change Erlang version in nerves?"

mattneel:
Seems the logical place. I'll add a section outlining the steps explicitly, as well as using the slash/search command in menuconfig to discover where in the menus a specific feature might live. I'll put together a PR soon.

PR #323, opened 2019-06-02
Update to Buildroot 2019.05

fhunleth:
This looks like a fairly boring Buildroot release. The Raspberry Pi firmware has been updated and we'll need to propagate that around. libp11 is now upstream.


fhunleth:
Update - I've tested libp11. It was built slightly differently from what I had done before, but it seems to work just fine.

PR #334, opened 2019-06-29
Raspberry Pi 4 updates

fhunleth:


fhunleth:
This PR is ready. It is probably advisable for us to update all Raspberry Pi systems to a newer version of the Linux kernel, but I suspect that it only is important for the Pi 4. Regardless, I like to have kernel version match across the systems we support and will make a pass the next chance that I get.

PR #363, opened 2019-09-05
Fix shellcheck docker-entrypoint warnings

mattludwigs:
```bash
λ  ~/code/nerves/nerves_system_br (master)  $ shellcheck support/docker/nerves_system_br/docker-entrypoint.sh 

In support/docker/nerves_system_br/docker-entrypoint.sh line 9:
  groupadd -o -g $GID nerves
                 ^-- SC2086: Double quote to prevent globbing and word splitting.


In support/docker/nerves_system_br/docker-entrypoint.sh line 10:
  useradd -o -g $GID -u $UID -m nerves
                ^-- SC2086: Double quote to prevent globbing and word splitting.


In support/docker/nerves_system_br/docker-entrypoint.sh line 14:
  gosu nerves $@
              ^-- SC2068: Double quote array expansions to avoid re-splitting elements.

```


fhunleth:
Wow, thanks for all of the shellcheck fixes!

Do you think it's possible for us to enable shellcheck on CI yet?

PR #366, opened 2019-09-10
Fix quote expansion

fhunleth:
A few shellcheck fixes broke wildcard expansions. This works around this
issues.


fhunleth:
Verified with nerves_system_bbb.

PR #369, opened 2019-09-16
Add python3-distutils to docker image

mobileoverlord:
Fixes build issues resulting in the error 
```
ModuleNotFoundError: No module named 'distutils.sysconfig'
configure: error: Python failed to run; see above error.
```


fhunleth:
Is it time to update the other Python 2 packages to their 3 versions?

PR #373, opened 2019-09-19
erlang: bump to OTP 22.1

fhunleth:
[From the release notes]

Erlang/OTP 22.1 is the first service release for the 22 major release
with new features, improvements as well as bugfixes

Potential Incompatibilities

* Mnesia: Transactions with sticky locks could with async_asym
  transactions be committed in the wrong order, since asym transactions
  are spawned on the remote nodes. To fix this bug the communication
  protocol between mnesia nodes had to be updated, thus mnesia will no
  longer be able to connect to nodes earlier than mnesia-4.14 , OTP-19.0.

* Stdlib: Debugging of time-outs in gen_statem has been improved.
  Starting a time-out is now logged in sys:log and sys:trace. Running
  time-outs are visible in server crash logs, and with sys:get_status. Due
  to this system events {start_timer, Action, State} and {insert_timout,
  Event, State} have been added, which may surprise tools that rely on the
  format of these events. New features: The EventContent of a running
  time-out can be updated with {TimeoutType, update, NewEventContent}.
  Running time-outs can be cancelled with {TimeoutType, cancel} which is
  more readable than using Time = infinity.{rel, Name, Vsn, RelApps,
  Opts}.

Highlights

Compiler:

* erlc can now automatically use a compile server to avoid starting an
  Erlang system for each file to be compiled in a multi-file project. See
  the documentation for how to enable it.

Standard libraries:

* SSL: Basic support for TLS 1.3 Client for experimental use. For more
  information see the Standards Compliance chapter of the User's Guide.

* crypto: The Message Authentication Codes (MAC) CMAC, HMAC and Poly1305
  are unified into common functions in the New Crypto API. See the manual
  for CRYPTO. cipher_info/1 functions returns maps with information about
  the hash or cipher in the argument.

For more details see
http://erlang.org/download/otp_src_22.1.readme

### DIFF
--- a/patches/buildroot/0011-erlang-support-OTP-20-21-and-22.patch
+++ b/patches/buildroot/0011-erlang-support-OTP-20-21-and-22.patch
@@ -1,4 +1,4 @@
-From b32a871b52f491b2d2c8532f9dd51b623e7666b5 Mon Sep 17 00:00:00 2001
+From 3b4800eb1c2b274b96a3a90ec64d96cd17e34696 Mon Sep 17 00:00:00 2001
 From: Frank Hunleth <fhunleth@troodon-software.com>
 Date: Tue, 11 Sep 2018 12:28:41 -0400
 Subject: [PATCH] erlang: support OTP 20, 21, and 22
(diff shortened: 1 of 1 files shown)


fhunleth:
Note: esl-erlang hasn't been updated to 22.1 yet

PR #380, opened 2019-11-01
fwup: bump to v1.4.0

fhunleth:
fwup v1.4.0 adds support for GPT partitions and makes it much easier to
handle UEFI-only boards.


fhunleth:
This one is intended to be merged after the Erlang one.

PR #390, opened 2019-11-26
nerves_initramfs: new package

fhunleth:
This pulls down a pre-built initramfs for modifying the boot of Nerves
devices. It makes it possible to revert bad builds automatically without
assistance from the primary bootloader (i.e., Raspberry Pi bootloader)
and also to mount the rootfs via the device mapper (like for encrypted
rootfs support). Note that secure key handling is not yet possible in
the pre-built package, but encrypted filesystems are mountable for test
purposes.


fhunleth:
Per conversation with Justin, let's have this hang around a little as a PR/branch to make testing easier. At some point, it sounds like we'll want to move `nerves_initramfs` to `nerves-project`. I'll update the PR to pull from `nerves-project` when that time comes. The only thought for merging earlier is to facilitate testing. Merging does not affect anyone's project since this has to be enabled and referenced by a device's bootloader for anything to happen. Even then, the default behavior for `nerves_initramfs` is to mount the rootfs and run `erlinit` which is identical to what happens now.

fhunleth:
Postponing this update so that it can incubate as a custom system's package.

PR #391, opened 2019-11-26
erlang: bump to 22.1.8

fhunleth:
From the release notes:

  OTP-16301    Application(s): erts
               Related Id(s): ERL-1079

               Large amounts of quickly executed dirty work could
               cause heavy contention on an internal spin lock. The
               spin lock was replaced by a mutex which behaves much
               better under these conditions.


mobileoverlord:
> quickly executed dirty work

Always brings trouble.

PR #402, opened 2019-12-18
Add Nerves-specific info to os-release

### LINKED ISSUE #401: Add Nerves version information to /etc/os-release
If logging into a system with an unknown version of firmware running on it, it's hard to figure out the Erlang, `nerves_system_br`, and `nerves_system_[device]` versions. Listing the `nerves_system_[device]` version could help determine all versions.

It looks like [/etc/os-release](https://www.freedesktop.org/software/systemd/man/os-release.html) could be updated to refer to Nerves rather than Buildroot and supply this information.

fhunleth:
The `/etc/os-release` file had contained information about Buildroot
that wasn't all that useful. This adds some information that could be
useful to know. Here's an example:

```
$ cat target/etc/os-release
NAME=Nerves
ID=nerves
NERVES_SYSTEM_BR_VERSION=1.10.0
NERVES_SYSTEM_NAME=nerves_system_rpi0
NERVES_SYSTEM_VERSION=0.14.0
NERVES_BUILDROOT_VERSION=2019.11
```

Fixes #401.


jjcarstens:
Is it worth adding versions for other things like full OTP and Elixir? Or the idea is those will just be implied based on the nerves system version you have?

fhunleth:
I wanted to, but...

The Nerves system doesn't supply the Elixir library so I can't add that.

I didn't know of an easy way to get the OTP version. If a simple way occurs to me, I'll send an update or PR. In the mean time, the best idea I have is to look at the release notes for the Nerves system that's referenced. It almost always has the OTP version listed in it.

PR #413, opened 2020-01-21
edk2: copy shell to $BINARIES_DIR

fhunleth:
This fixes an issue where the `efi-part` directory hasn't been created
so the shell gets renamed as `efi-part`. Since `shell.efi` doesn't get
put in the root of the EFI system partition anyway, I moved it to the
root of the images directory and scripts can put it where ever makes
sense.


mobileoverlord:
:facepalm:

PR #420, opened 2020-02-14
Apply system-specific buildroot patches

mobileoverlord:
Apply any buildroot patches located in `$NERVES_DEFCONFIG_DIR/patches/buildroot` if the directory exists. These patches are applied following the nerves patches.


fhunleth:
I'm not sure I like this patch. What problem is it trying to solve? I'm pretty sure there's a better way.

mobileoverlord:
Sorry, I realize now that the better way to apply these patches is through `BR2_GLOBAL_PATCH_DIR` and to use the project specific patches docs found in `9.8. Adding project-specific patches` of the [buildroot docs](https://buildroot.org/downloads/manual/manual.html)

PR #438, opened 2020-04-13
Update erlinit project repo url

celaxodon:
I just noticed an out of date repository URL when looking at buildroot integration.


fhunleth:
Good catch. Thanks!

PR #450, opened 2020-05-08
erlang: bump to 22.3.4

fhunleth:
See the following release notes for changes:

* https://erlang.org/download/OTP-22.3.4.README

The disksup patch has been included in the main OTP version bump patch
for ease of maintenance.


fhunleth:
I probably should have marked this as draft. We really need to wait for ESL to push an Erlang 23.0 package for docker like #450 is currently doing for that version.

PR #453, opened 2020-05-13
erlang: bump to 23.0

fhunleth:
This is a major bump to Erlang/OTP and contains many changes. See
the release note at https://erlang.org/download/OTP-23.0.README.


fhunleth:
I probably should have marked this as draft. We really need to wait for ESL to push an Erlang 23.0 package for docker like #450 is currently doing for that version.

fhunleth:
Wow, looks like ESL posted the updated Erlang package, so this is good to go. 

As a note, it's still possible to use OTP 22, but it has to be manually selected now.

PR #454, opened 2020-05-14
Buildroot: Bump to 2020.05-rc1

fhunleth:
This is a major release. From the release notes:

http://lists.busybox.net/pipermail/buildroot/2020-May/282192.html


fhunleth:
This is the 2020.05-rc1 release at the moment. It has a nice update to Qt that lets us finally drop our Qt patches. I haven't tested it on the kiosk systems, though.

fhunleth:
Everything builds with this update except kiosk_system_x86_64, but it hasn't built in a while. 

I'd like to make an rc release of `nerves_system_br` so that it's easier for us to test out the new Buildroot and Erlang 23.0. The idea would be that as soon as Buildroot officially releases 2020.05, we'd make a real release of `nerves_system_br`.

PR #456, opened 2020-05-18
Remove duplicate package

mattludwigs:
Looks like this package is already in the package list here: https://github.com/mattludwigs/nerves_system_br/blob/master/support/docker/nerves_system_br/Dockerfile#L36. I could be missing some detail though.


fhunleth:
Thanks!

PR #469, opened 2020-07-21
erlang: bump to 23.0.3

fhunleth:
This bumps Erlang from 23.0.2 to 23.0.3. Changes can be found at
https://erlang.org/download/OTP-23.0.3.README.


fhunleth:
Note: Erlang version in DockerFile not bumped since ESL hasn't posted an update yet.

PR #473, opened 2020-08-15
Fix up and start running the smoke test again

fhunleth:
The unit tests were forgotten about a long time ago. This combines the
two tests and updates it to be a better smoke test of
`nerves_system_br`.


fhunleth:
@mobileoverlord Help? Is it possible to build the Docker image and then use it to run the smoke test? That seems like that would be the best way of doing this. As you can see, I'm currently running the smoke test in the alpine container, but that's because I didn't know how to do test in the container that was being built. Running in the alpine container won't work for many reasons, but I wanted to get something started.

fhunleth:
Hmm... I updated to test against `latest`, so it's still not right, but maybe it's slightly better.

jjcarstens (approved):

mobileoverlord (approved):
I updated the circle script to always build an image and push it to the remote as `nervesproject/nerves_system_br:$CIRCLE_SHA1` to use for the tests.

.circleci/config.yml
```diff
@@ -44,25 +43,27 @@ docker_save: &docker_save
...
 docker_push: &docker_push
   run:
     name: Push docker images to dockerhub
     command: |
       docker login -u "$DOCKER_USER" -p "$DOCKER_PASS"
-      docker push $DOCKER_REPO:$DOCKER_TAG
-      docker push $DOCKER_REPO:latest
+
+      if [ -z "$CIRCLE_TAG" ]; then
+        docker push $DOCKER_REPO:$DOCKER_TAG
```
fhunleth, line 61: Should we be setting some TTL on these images to be good citizens?

.circleci/config.yml
```diff
@@ -71,7 +77,15 @@ jobs:
       - save_cache:
           key: docker-{{ .Branch }}-{{ .Revision }}
           paths: "/docker"
-      
+
+  test:
+    docker:
+      - image: nervesproject/nerves_system_br:$CIRCLE_SHA1
```
fhunleth, line 83: Won't this fail for builds where `$DOCKER_TAG` is the tagged version?

mobileoverlord:
I updated this to only push to the tag `dev` and to push to that tag on every build. If the build is tagged, it will also push to `latest` and `<VERSION>`

fhunleth:
Thanks! I'm going to squash a couple commits and then merge!

PR #492, opened 2020-10-18
rpi-firmware: move version maintenance to system

fhunleth:
This moves the rpi-firmware version selection to the Nerves system. This
makes it much easier to update kernels since you no longer need to
update `nerves_system_br` to add an option to select the new
rpi-firmware.

All Raspberry Pi Nerves systems will need to be updated to add something
like the following line to their `nerves_defconfig`:

```
BR2_PACKAGE_RPI_FIRMWARE_CUSTOM_VERSION="1.20200902"
```

The default is to use the Buildroot-supplied version.


fhunleth:
This change has a downside - it puts the onus on the Nerves system maintainer for matching rpi-firmware to their Linux kernel. The upside, though, is that it makes it way easier to see that you have matched versions if you use the Raspberry Pi Foundations tags (like `1.20200902`).

ConnorRigby (approved):
This is interesting and i never would have thought of doing this. Looks good to me

mobileoverlord:
🎉

mobileoverlord (approved):

fhunleth:
FYI - I updated this to give an error if the firmware version isn't set. I didn't feel good about having a default. If you have the wrong version, device tree things sometimes don't work and that seem hard to debug. Providing a default seems like an easy way to mess someone up.

PR #500, opened 2020-11-30
rpi-userland: support aarch64

fhunleth:
This makes it possible to build rpi-userland on aarch64. Enabling rpi-userland with this patch will fix https://github.com/elixir-circuits/circuits_gpio/issues/84.


fhunleth:
I updated this to support building the GPU drivers. These come from `mesa3d`.

mobileoverlord (approved):

PR #508, opened 2020-12-18
erlang: bump version to 23.2.1

fhunleth:
This bumps Erlang from 23.1.5 to 23.2. See
https://erlang.org/download/OTP-23.2.README

PR #509, opened 2020-12-27
Bump docker/tool-versions to Erlang 23.1.5

fhunleth:


fhunleth:
Yes, this seems weird since #508 bumps Erlang to 23.2.1. Given that 23.2.1 has the backwards-incompatible SSL changes, I'm uneasy about merging 508 quickly. Perhaps #508 gets merged to `main` and is released with gcc-10.

jjcarstens:
👌 I'm on board with being better safe than sorry

jjcarstens (approved):

PR #529, opened 2021-02-20
Set CPPFLAGS for consistency

fhunleth:
Nerves already sets CFLAGS and CXXFLAGS. It hadn't set CPPFLAGS, but it
really should set that as well to avoid anything from the host being
incorrectly used.


mobileoverlord (approved):
I think I just started to bump into this.

PR #534, opened 2021-03-22
Allow the ability to not scrub directories in the release dir

ConnorRigby:
To exercise this, one should create a `.noscrub` file in any
directory inside the release. any files in the same directory
will be excluded from scrubbing

### DIFF
--- a/scripts/scrub-otp-release.sh
+++ b/scripts/scrub-otp-release.sh
@@ -105,7 +105,14 @@ get_expected_executable_type()
     rm "$tmpfile"
 }
 
-EXECUTABLES=$(find "$RELEASE_DIR" -type f -perm -100)
+# searches for a file named `.noscrub`. If found
+# filter that folder out of the `EXECUTABLES` list
+NOSCRUBS=$(find "$RELEASE_DIR" -name .noscrub)
+for NOSCRUB in $NOSCRUBS; do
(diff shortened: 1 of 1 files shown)


fhunleth:
I like how simple this is. My only comment is that it needs to be documented so that people can discover it. IMHO, the best place I can think of is in the error message right below your change. Not sure if you can think of anything better.

ConnorRigby:
> I like how simple this is. My only comment is that it needs to be documented so that people can discover it. IMHO, the best place I can think of is in the error message right below your change. Not sure if you can think of anything better.

Yeah i sort of started making an addition to the "customizing your system" doc, but it didn't really belong there. I like your idea of the error message.

ConnorRigby:
@fhunleth just pushed another update with the error message update.

scripts/scrub-otp-release.sh
```diff
@@ -144,6 +151,10 @@ for EXECUTABLE in $EXECUTABLES; do
             echo "4. Are you building outside of Nerves' mix integration? If so,"
             echo "   make sure that you've sourced 'nerves-env.sh'."
             echo
+            echo "If you are very sure you know what you are doing, you may place an empty"
+            echo "file in the same directory as the offending file(s) called `.noscrub`. "
+            echo "This will explitly disable scrbbing for that directory."
```
fhunleth, line 156:
```suggestion
            echo "This will explicitly disable scrubbing for that directory."
```

fhunleth (approved):

PR #537, opened 2021-03-23
Fix scrub-otp-release script for when `.noscrub` isn't used

ConnorRigby:


entone:
I couldn't help myself.

no scrubs

ConnorRigby:
@entone i've been fighting the urge for 2 days!

entone:
You're a stronger person than I am :)

fhunleth:
@ConnorRigby I reworked the script and added code to exercise it to the smoke test. The smoke test doesn't check the no .noscrub case. My excuse is that I'm lazy and that since that case is so common, it will be hard to miss. I'll manually verify that shortly.

tests/smoke-test/rootfs_overlay/srv/erlang/erts-11.1.8/bin/start
```diff
@@ -0,0 +1,37 @@
...
+# %CopyrightEnd%
+#
+# This program invokes the erlang emulator by calling run_erl.
+# It should only be used at an embedded target system.
+# It should be modified to give the correct flags to erl (via start_erl),
+# e.g -mode embedded -sname XXX
+#
+# Usage: start [Data]
+#
+ROOTDIR=/home/fhunleth/.asdf/installs/erlang/23.2.7
```
jjcarstens, line 28: Seems like this shouldn't work in CI?

jjcarstens (approved):

jjcarstens:
Sorry...meant to hit "comment", but actually hit "approve" - though I think it should be okay

tests/smoke-test/rootfs_overlay/srv/erlang/erts-11.1.8/bin/start
fhunleth: Interesting. That file should never be run and should be deleted by the scrubber.

PR #542, opened 2021-04-23
Fix SC2086 of board/nerves-common/post-*.sh

pojiro:
https://github.com/koalaman/shellcheck/wiki/SC2086


fhunleth:
Looks good. Thanks!

PR #543, opened 2021-04-23
Fix SC2115 of tests/run_tests.sh

pojiro:
https://github.com/koalaman/shellcheck/wiki/SC2115


fhunleth:
Nice! Thanks for adding the `:?` part. Sorry for taking so long to look at it.

PR #546, opened 2021-05-08
erlang: bump version to 23.3.3

fhunleth:
This bumps Erlang from 23.3.2 to 23.3.3. See
https://erlang.org/download/OTP-23.3.3.README.


fhunleth:
See https://github.com/erlang/otp/issues/4813

PR #561, opened 2021-07-03
Allow buildroot patches from $NERVES_DEFCONFIG_DIR/patches/buildroot

erauer:
Allow additional buildroot patches for custom nerves systems

### DIFF
--- a/create-build.sh
+++ b/create-build.sh
@@ -121,6 +121,10 @@ create_buildroot_dir() {
     # Apply Nerves-specific patches
     "$NERVES_SYSTEM/buildroot/support/scripts/apply-patches.sh" "$NERVES_SYSTEM/buildroot" "$NERVES_SYSTEM/patches/buildroot"
 
+     if test -d "$NERVES_DEFCONFIG_DIR/patches/buildroot"; then
+    "$NERVES_SYSTEM/buildroot/support/scripts/apply-patches.sh" "$NERVES_SYSTEM/buildroot" "$NERVES_DEFCONFIG_DIR/patches/buildroot"
+     fi
+
     # Symlink Buildroot's dl directory so that it can be cached between builds
     ln -sf "$NERVES_BR_DL_DIR" "$NERVES_SYSTEM/buildroot/dl"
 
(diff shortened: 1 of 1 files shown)


fhunleth (commented):
I'm going to need to think about this. I understand why it's desired. 

It won't work for anyone on the core team since we all build using a shared `nerves_system_br` directory for all systems rather than one per system. I also would never accept a Buildroot patch in any of the official systems since I'd want patches to either be Nerves-specific and be submitted to `nerves_system_br`, or general and be submitted to Buildroot. However, I know that most people don't submit patches upstream, and when you're in a time crunch or experimenting it's really annoying to have to fork everything for a trivial change.

elcritch:
This would be handy for webengine build for rpi4's. It'd be nice to set it up to follow upstream by rebasing and rebuilding. The patch I copied was initially proposed upstream but never got merged (though mainly for lack of follow up from the person proposing it). There could be cases where perhaps patches upstream aren't accepted for various reasons. 

One thing to point out is that I've never _not_ had to patch buildroot to get the more complex browsers to compile. :face_with_head_bandage: so the use case of experimenting is likely to be useful in the future.

erauer:
@fhunleth I think you articulated the problems and advantages to this mechanism better than I could. I had considered whether this should include documentation and deferred in part since it could be overly abused. The `Adding a custom Buildroot Package`  definitely articulated the philosophy for committing upstream, https://hexdocs.pm/nerves/customizing-systems.html .

fhunleth:
It's been a couple weeks. I don't know if any patches have been submitted upstream with the kiosk efforts. I hope so. That should hopefully resolve whatever caused this need in the first place. 

The main problem I have with this patch is that any system that uses this patch can't be supported by the core team. I just never build systems in this way except in the CI scripts.

My recommendation is to submit Buildroot patches to the Buildroot mailing list. If you need them before they can officially be released in Buildroot and it's important that they be in `nerves_system_br`, I will take a look at pulling in the unreleased Buildroot commits temporarily as patches in `nerves_system_br`.

Another option is to copy/paste the Buildroot recipe that you want to patch into its own directory.

PR #563, opened 2021-07-11
Allow systems to be built on ARM 64 Linux

mobileoverlord:
Running on ARM 64 bit Linux I am getting the error `ERROR: 64-bit Linux required for running cross-compilers built by nerves-toolchain`. Updates the check to include ARM 64 linux.

### DIFF
--- a/create-build.sh
+++ b/create-build.sh
@@ -62,7 +62,7 @@ if [[ $HOST_OS != "Linux" ]]; then
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


fhunleth:
This is probably going to cause some pain down the line if aarch64 `nerves_defconfig` files ever become common. I suppose this will become an issue for Buildroot people too. 

I don't see any reason to not allow it. Perhaps at some point we should just delete the check.

PR #565, opened 2021-07-17
Optionally include SBOM info in system artifacts

fhunleth:
Buildroot has two tools for capturing SBOM-relevant information: `make
pkg-stats` and `make legal-info`. This makes it possible to add
`pkg-stats` and `legal-info` targets to the `make_opts` for your system
so that this info is captured. This is convenient so that you don't have
to rebuild a system to get the files when you need them. It also makes
it possible for tools to parse them and combine them with SBOM data from
the rest of your project.

The increase in system artifact size is minimal. Buildroot also includes
all of the source tarballs in its legal info, but that's like 800+ MB
for projects, so that's not included in this change.


ConnorRigby (approved):

jjcarstens (approved):
LGTM

PR #573, opened 2021-07-26
Fix scrubbing files on MacOS

jjcarstens:
For whatever reason, the `find` command doesn't expand the exclusions on MacOS and so if you are attempting to skip scrubbing files, the task will fail. To get the proper variable expanded, we need to `eval` the call.

But adding `eval` on any other system breaks the script. So, check the OS and decide which to do first.

I have no idea why MacOS won't expand and the only hints on the internet were to `eval` to force expansions first. I've confirmed this works in MacOS and Linux


ConnorRigby (approved):
Didn't test, but it makes sense to me. great find

fhunleth (approved):

PR #578, opened 2021-08-29
Fix /bin/bash shebang to be /usr/bin/env bash

mayl:
This wasn't a pull request initially because it took a minute for me to determine where this file came from on GitHub.  Using `/usr/bin/env` seems consistent with the other scripts here so opening as a pull request.

resolves [#650](https://github.com/nerves-project/nerves/issues/650)

### DIFF
--- a/board/nerves-common/post-build.sh
+++ b/board/nerves-common/post-build.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 set -e
 
--- a/scripts/ci-deploy.sh
+++ b/scripts/ci-deploy.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 # Collect the build artifacts and put them in a directory
 # for Travis-CI to publish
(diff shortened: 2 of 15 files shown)


mayl:
I also noted that some of these scripts reference `sh` while others use `bash`.  Is there something intentional there or can/should they be condensed down to all just use `bash`?

fhunleth:
Thank you for sending this PR! It has been a really long time since I've thought about shebangs, and I agree that calling `env` is the way to go. As for `bash` vs `sh`, the original scripts only required `sh` and then a bash-ism crept in, so we started switching to `bash`. Making them consistently invoke `bash` seems like a good idea.

Want to make the `sh` -> `bash` change and then I'll merge?

mayl:
No problem, thanks for reviewing!

Based on your input, I went ahead and audited the rest of the scripts and found some more with hardcoded paths and/or references to `sh` instead of `bash`.  I've moved those all to `/usr/bin/env bash` and squashed down to one commit for you in this PR. Let me know if this looks OK to you still.

fhunleth:
This is perfect. Thank you!!!

PR #579, opened 2021-08-30
Refresh Busybox and turn off libgcc static linking

fhunleth:
Some options got turned on during Busybox updates. This turns them back off.


jjcarstens (approved):

fhunleth:
I removed the `libgcc` commit. It turns out that I was accidentally comparing a stripped to unstripped build of busybox. The libgcc option didn't change the size so I left the default.

PR #582, opened 2021-09-03
erlang: bump to 24.0.6

fhunleth:
This bumps Erlang from 24.0.5 to 24.0.6. See
https://erlang.org/download/OTP-24.0.6.README.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,1 +1,1 @@
-erlang 24.0.5
+erlang 24.0.6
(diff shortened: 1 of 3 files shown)


jjcarstens (approved):
I actually see the buffer overflow problem occasionally with CTRL+R, so this will be nice 👍

fhunleth:
I think that I've seen the buffer overflow problem too!!!

PR #603, opened 2021-11-28
fwup: work around GitHub URL change

fhunleth:
This is a temporary patch to work around the GitHub URL change which
brok the download. Once we update to a version of Buildroot with the
GitHub URL fix, this patch should be removed.


fhunleth:
Not needed since GitHub reverted the change that broke downloads.

PR #616, opened 2022-01-01
1.18 changelog improvement

axelson:
Hopefully make it a little easier for people to understand

Question: What does "extended firmware" mean? Is that just if you have your own `fwup.conf` that you're using?


fhunleth:
This looks great. Thanks!

PR #618, opened 2022-01-14
export `CMAKE_TOOLCHAIN_FILE` for downstream libraries which use CMake

cocoa-xu:
Hi, this PR adds a basic CMake toolchain file for cross-compiling environment and the path to the toolchain file is exported to a system environment variable `CMAKE_TOOLCHAIN_FILE`. 

This allows downstream libraries which use CMake to find the correct sysroot, and prevent CMake to search system libraries in the host environment.

The downstream libraries can decide whether they want to use the provided toolchain file. If yes, then they can append `-D CMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}` to their CMake options.

For example, before applying this patch, CMake will search libraries available on the host, which causes the linker to try to link the NIF shared library to libraries on the host, and this can result in either

- a fatal error, when the host's architecture is different from the `MIX_TARGET`
- successfully compiled (when host and `MIX_TARGET` have the same CPU architecture), but the libraries linked were never copied to the firmware, the final program cannot load the library and lead to infinite reboots

#### before
```
==> nerves
==> nerves_project

Nerves environment
  MIX_TARGET:   rpi0
  MIX_ENV:      dev

... skipped ...

--   Media I/O:
--     ZLib:                        zlib (ver 1.2.11) <= library on x86_64 host, yet the target is armv6
--     JPEG:                        libjpeg-turbo (ver 2.1.0-62)
--     WEBP:                        /usr/local/lib/libwebp.a (ver encoder: 0x0210) <= library on x86_64 host, yet the target is armv6
--     PNG:                         build (ver 1.6.37)
--     TIFF:                        build (ver 42 - 4.2.0)
--     JPEG 2000:                   OpenJPEG (ver 2.4.0)
--     HDR:                         YES
--     SUNRASTER:                   YES
--     PXM:                         YES
--     PFM:                         YES
...
```

#### after
```
==> nerves
==> nerves_project

Nerves environment
  MIX_TARGET:   rpi0
  MIX_ENV:      dev

... skipped ...

--   Media I/O:
--     ZLib:                        /Users/cocoa/.nerves/artifacts/nerves_system_rpi0-portable-1.18.0/staging/usr/lib/libz.so (ver 1.2.11) <= use library in sysroot, ok
--     JPEG:                        libjpeg-turbo (ver 2.1.0-62)
--     WEBP:                        build (ver encoder: 0x020f) <= build library for armv6, ok
--     PNG:                         build (ver 1.6.37)
--     TIFF:                        build (ver 42 - 4.2.0)
--     JPEG 2000:                   OpenJPEG (ver 2.4.0)
--     HDR:                         YES
--     SUNRASTER:                   YES
--     PXM:                         YES
--     PFM:                         YES
...
```

Signed-off-by: Cocoa <89497197+cocoa-xu@users.noreply.github.com>

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -39,6 +39,7 @@ defmodule Nerves.System.BR.Mixfile do
         "mix.exs",
         "nerves_env.exs",
         "nerves-env.sh",
+        "nerves-env.cmake",
         "nerves.mk",
         "README.md",
         "VERSION"
--- a/nerves-env.cmake
+++ b/nerves-env.cmake
@@ -1,5 +1,3 @@
-cmake_minimum_required(VERSION 2.9 FATAL_ERROR)
-
 set(CMAKE_SYSTEM "Linux")
 set(CMAKE_SYSTEM_NAME "Linux")
 
--- a/nerves-env.cmake
+++ b/nerves-env.cmake
@@ -1,0 +1,15 @@
+cmake_minimum_required(VERSION 2.9 FATAL_ERROR)
+
+set(CMAKE_SYSTEM "Linux")
+set(CMAKE_SYSTEM_NAME "Linux")
+
+# set nerves sysroot
+set(CMAKE_FIND_ROOT_PATH "$ENV{NERVES_SDK_SYSROOT}")
+
+# adjust the default behavior of the find commands:
+# search headers and libraries in the target environment
+set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
+set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
+
+# never search programs in the host environment
+set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
--- a/nerves-env.cmake
+++ b/nerves-env.cmake
@@ -2,7 +2,7 @@ set(CMAKE_SYSTEM "Linux")
 set(CMAKE_SYSTEM_NAME "Linux")
 
 # set nerves sysroot
-set(CMAKE_FIND_ROOT_PATH "$ENV{NERVES_SDK_SYSROOT}")
+set(CMAKE_FIND_ROOT_PATH "$ENV{NERVES_SDK_ROOTFS}")
 
 # adjust the default behavior of the find commands:
 # search headers and libraries in the target environment
--- a/nerves-env.cmake
+++ b/nerves-env.cmake
@@ -2,7 +2,7 @@ set(CMAKE_SYSTEM "Linux")
 set(CMAKE_SYSTEM_NAME "Linux")
 
 # set nerves sysroot
-set(CMAKE_FIND_ROOT_PATH "$ENV{NERVES_SDK_ROOTFS}")
+set(CMAKE_FIND_ROOT_PATH "$ENV{NERVES_SDK_SYSROOT}")
 
 # adjust the default behavior of the find commands:
 # search headers and libraries in the target environment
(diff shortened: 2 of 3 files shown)


fhunleth:
Nice! I'm planning on making a Nerves system releases over the next few days or early next week. I'll add this to my queue to try out and include. It looks good based on inspection.

cocoa-xu:
I just found that `$NERVES_SDK_SYSROOT` is not quite what we need here. The real sysroot should be the unpacked version of the `~/.nerves/artifacts/nerves_system_${MIX_TARGET}-portable-${VERISON}/images/rootfs.squashfs`.

Can we have an unarchived version of this file in, say `~/.nerves/artifacts/nerves_system_${MIX_TARGET}-portable-${VERISON}/images/rootfs`?

cocoa-xu:
If we ignore the changes in the unarchived rootfs directory, then we only need to add a few lines to `unsquashfs` the `rootfs.squashfs` during installation and no changes for all other scripts.

Or, we can use the unarchived rootfs directory as the base, and overlay other things, e.g., the Erlang/OTP release, onto it. Something like

```bash
cd ~/.nerves/artifacts/nerves_system_${MIX_TARGET}-portable-${VERISON}/images
cp -a rootfs working_copy
# overlay the Erlang/OTP release to working_copy
# mksquashfs working_copy ...
```

fhunleth:
The `rootfs.squashfs` shouldn't be unpacked. The SDK sysroot contents are what compilers should reference. 

I think that you ran into a feature or a bug. What exactly did you run into?

If it has to do with OpenMP, there's a fix in a PR to all of the systems. I checked it out earlier today and it works, so it will be merged depending on when you look for it.

cocoa-xu:
But based on my understanding, the contents in `rootfs.squashfs` (along with overlays) will be built into the firmware, and that means libraries in `${NERVES_SDK_SYSROOT}` will not be copied. 

Therefore, if we set the `CMAKE_FIND_ROOT_PATH ` to `${NERVES_SDK_SYSROOT}`, and link a shared library in `${NERVES_SDK_SYSROOT}` but not in `rootfs.squashfs`, then in that case, when we try to boot the device using this firmware, `ld` won't be able to load the linked library (as it is not in the image/firmware) and result in infinite reboots, right?

fhunleth:
Right, but there are other differences. If there's a discrepancy between the rootfs.squashfs and the sysroot, then the sysroot needs to be updated. Unpacking the rootfs.squashfs will result in other issues.

cocoa-xu:
I was comparing the contents in both directories, and I found that we have `libSegFault.so` in the `nerves_system_rpi4-portable-1.18.0`, but it's in the `rootfs.squashfs`. So I thought it might be safe to use the `rootfs.squashfs`. Not quite sure if this (missing `libSegFault.so` in the rootfs) was intended or something else...

fhunleth:
Ah. The toolchain provided shared libraries are untracked by Buildroot and have to be specified a special way. That was the issue with OpenMP (I.e., `libgomp.so` was provided in the toolchain, but not copied). It is really unfortunate that it exists in the sysroot. It shouldn't be in sysroot. 

Having said that, `libSegFault.so` is 9K stripped, so if you want to use it, I'd be in favor of copying it over to the sysroot in the next release. This is trivial and I'll make the PR.

fhunleth:
This `cmake` patch works great in my tests. Thanks!

cocoa-xu:
Oh, I see! It should be fine; I was worrying if I set the `CMAKE_FIND_ROOT_PATH` to a wrong directory!

PR #625, opened 2022-02-06
Expose the gcc args that Buildroot uses

fhunleth:
This will make sure that they're always included in archives. The idea
is that if there's ever any compile flag confusion, we can consult the
Nerves system archive to see what Buildroot used when building.

This will also be useful with crafting the `TARGET_GCC_FLAGS` in the
system's `mix.exs`.


fhunleth:
See #625 and https://github.com/nerves-project/nerves/pull/698.

jjcarstens (approved):
Assuming this should pass once the others are merged

fhunleth:
It failed because I forgot to add a file to the commit. It's fixed now, so it should pass now. 🤞

jjcarstens (approved):

PR #627, opened 2022-02-11
Bump Erlang versions for Docker image

fhunleth:
When updating to OTP 24.2.1, these updates were missed. Luckily, it
doesn't look like there's a meaningful different with 24.2, but they
should match anyway.

### DIFF
--- a/support/docker/nerves_system_br/Dockerfile
+++ b/support/docker/nerves_system_br/Dockerfile
@@ -1,4 +1,4 @@
-FROM hexpm/erlang:24.2-ubuntu-focal-20210325
+FROM hexpm/erlang:24.2.1-ubuntu-focal-20210325
 LABEL maintainer="Nerves Project developers <nerves@nerves-project.org>" \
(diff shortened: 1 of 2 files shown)


support/docker/nerves_system_br/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM hexpm/erlang:24.2-ubuntu-focal-20210325
+FROM hexpm/erlang:24.2.1-ubuntu-focal-20210325
```
jjcarstens, line 1:
```suggestion
FROM hexpm/erlang:24.2.1-ubuntu-impish-20211102
```
Is there a particular reason to stay on `focal` vs the newer `impish`? If so, then `24.2.1-ubuntu-focal-20211006` is the tag you'll need here

jjcarstens: O good point. I totally didn't notice it wasn't LTS

PR #634, opened 2022-03-14
v1.18.6 release

jjcarstens:


ConnorRigby (approved):

fhunleth (approved):

fhunleth:
Before we ship this I think we should upgrade Erlang to 24.3.1 to fix the SSL issue with 24.3.

fhunleth (approved):

PR #642, opened 2022-05-13
Update RPi WiFi to 20210315-3+rpt5 + Zero 2W fix

fhunleth:
From the changelog:

* Add ability to choose between 43436 and 43436S using update-alternatives
* The shipping firmware for the SYN43436P does not support 4-way
  handshake offloading. This new firmware (version string "Version:
  9.88.4.77 CRC: 143f9f15 Date: Thu 2022-03-31 17:25:16 CST Ucode Ver:
  1043.20743 FWID: 01-3b307371") fixes that.


jjcarstens (approved):

fhunleth:
I haven't tested this out enough to feel comfortable merging. It seems like it should be tried on a Raspberry Pi Zero 2W first.

fhunleth:
Seems to work for me.

PR #644, opened 2022-05-18
buildroot: bump to 2022.02.1

fhunleth:
This extends the OTP-25 PR. My testing so far has been in this configuration and
it appears to work well.


IvoBCD:
> Why don't we need the fwup patch anymore?

@jjcarstens Buildroot 2022.02.1 already specifies fwup v1.9.0 (https://github.com/buildroot/buildroot/blob/9da87c52934842026918b9eb8511b35cd935c552/package/fwup/fwup.mk), so the fwup.mk file no longer needs to be patched to update to v1.9.0

fhunleth:
I just added a commit to fix a minor build issue for x86_64 systems. All official systems and several other systems build for me now.

fhunleth:
Update: I just added a commit that fixes an issue where Nerves heart is not installed correctly. The commit also makes the compilation step break now so you don't have to run the build to see the issue.

At this point, this is working on several platforms in my tests.

PR #656, opened 2022-06-20
set system processor

cocoa-xu:
Set `CMAKE_SYSTEM_PROCESSOR` to `$ENV{TARGET_ARCH}`. 

Some libraries (e.g., xnnpack) (implicitly) require this variable to be set when cross-compiling.


fhunleth:
Thanks!

PR #660, opened 2022-06-25
Use --sysroot instead of -I to include system headers

fhunleth:
The `--sysroot` does more than just include the system headers. Most
importantly, it turns off quite a few error messages with compiling
system header files. Here's an example error that's fixed:

```
/Users/fhunleth/.nerves/artifacts/nerves_system_mangopi_mq_pro-portable-0.1.0/staging/usr/include/stdio.h:28:5: error: "__cplusplus" is not defined, evaluates to 0 [-Werror=undef]
   28 | #if __cplusplus >= 201103L
      |     ^~~~~~~~~~~
cc1: some warnings being treated as errors
```

While `--sysroot` should have been used in the beginning, it wasn't
until the RISC-V port that this showed up. I think that it's due to the
combination of the gcc 11.3 update, the use of the Musl C library AND
the fact that `nerves_key_pkcs` has `-Werror=undef` in its options. It's
possible that warnings were printed but ignored in the past. That final
option to make warnings fail the build highlighted the issue.


scripts/nerves-env-helper.sh
```diff
@@ -110,9 +110,9 @@ export CROSSCOMPILE
 export REBAR_PLT_DIR=$NERVES_SDK_SYSROOT/usr/lib/erlang
 export CC=$CROSSCOMPILE-gcc
 export CXX=$CROSSCOMPILE-g++
-export CFLAGS="-D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64  -pipe -O2 -I$NERVES_SDK_SYSROOT/usr/include"
-export CPPFLAGS="-D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -I$NERVES_SDK_SYSROOT/usr/include"
-export CXXFLAGS="-D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64  -pipe -O2 -I$NERVES_SDK_SYSROOT/usr/include"
+export CFLAGS="-D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64  -pipe -O2 --sysroot $NERVES_SDK_SYSROOT/usr/include"
```
jjcarstens, line 113:
```suggestion
export CFLAGS="-D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64  -pipe -O2 --sysroot $NERVES_SDK_SYSROOT"
```
Shouldn't just be the sysroot path to match what is happening in the Elixir file as well?

fhunleth: Yes, omg, thanks for the catch.

jjcarstens (approved):

PR #663, opened 2022-07-19
Enable aarch64 compatibility in the Docker image

doawoo:
Currently I develop primarily on an M1 MacBook Pro - When it comes to building Nerves systems, it seems that there's only first class support for `x86_64`. 

This PR aims to introduce small changes to the Dockerfile and CI configuration that will enable `aarch64` host systems (such as M1 machines) to build Nerves systems using Docker. I've tested it on some internal system repositories with success.

* Enable multi-arch builds in CircleCI
* Modify `Dockerfile` to work with both targets (`linux/amd64` and `linux/arm64`)
* Add a `setup-packages.sh` script used during the Docker build to use the proper method of installing `i386` libraries for systems.

NOTE: To fully support `aarch64` host machines, we also need to enable `aarch64` Linux toolchains over at: https://github.com/nerves-project/toolchains in the releases. I've used Docker to build these successfully as well.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,8 @@ docker_build: &docker_build
   run:
     name: Build docker images
     command: |
-      docker build -t $DOCKER_REPO:$DOCKER_TAG $DOCKER_PATH
+      docker buildx create --use
+      docker buildx build --platform linux/arm64,linux/amd64 $DOCKER_REPO:$DOCKER_TAG $DOCKER_PATH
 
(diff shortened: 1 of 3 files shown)


fhunleth:
This is great! Thank you!

I'll work with you to get aarch64 toolchains. That was an omission on the last release due to lack of time and it not being automated. I'm very glad that you're helping fix this.

fhunleth (approved):

fhunleth:
Thanks!

Issue #202, opened 2018-08-02
Fix shellcheck warnings on all shell scripts


fhunleth:
This project has quite a few shell scripts that are missing double quotes around variables and other issues. It would be great if they could all be fixed since it is highly likely that they're causing problems with paths that contain spaces.

Since it looks like there are quite a few warnings, PRs addressing small pieces (like a file or one type of warning) at a time would be appreciated.

pojiro:
I'd like to try to fix this, #202. Tag [v1.15.1](https://github.com/nerves-project/nerves_system_br/tree/v1.15.1) has below shell scripts, and current status is below. I will create PR.

|`find nerves_system_br/ -name "*.sh"`|shellcheck|link to PR|
|-|-|-|
|nerves_system_br/tests/smoke-test/verify.sh|- (no warnings)|- (nothing to do)|
|nerves_system_br/tests/smoke-test/rootfs_overlay/srv/erlang/releases/0.1.0/env.sh|-|-|
|nerves_system_br/tests/smoke-test/run_scrubber.sh|-|-|
|nerves_system_br/tests/run_tests.sh|SC2115||
|nerves_system_br/scripts/scrub-target.sh|-|-|
|nerves_system_br/scripts/mksystem.sh|-|-|
|nerves_system_br/scripts/download-buildroot.sh|-|-|
|nerves_system_br/scripts/rel2fw.sh|-|-|
|nerves_system_br/scripts/ci-deploy.sh|-|-|
|nerves_system_br/scripts/scrub-otp-release.sh|-|-|
|nerves_system_br/scripts/buildroot-state.sh|-|-|
|nerves_system_br/scripts/ci-build.sh|-|-|
|nerves_system_br/scripts/nerves-env-helper.sh|-|-|
|nerves_system_br/scripts/mksyslinuxfs.sh|-|-|
|nerves_system_br/nerves-env.sh|muliple SC warnings||
|nerves_system_br/board/nerves-common/post-createfs.sh|SC2086||
|nerves_system_br/board/nerves-common/post-build.sh|SC2086||
|nerves_system_br/create-build.sh|-|-|
|nerves_system_br/support/docker/nerves_system_br/docker-entrypoint.sh|-|-|

pojiro:
I'm not the right person to fix nerves-env.sh because the script is  a little bit complicated and I don't understand the script's intention.

PR #675, opened 2022-09-22
Fixup shellchecks for nerves-env.sh

jjcarstens:
For #202 

* https://www.shellcheck.net/wiki/SC2046 - Quote this to prevent word splitting.
* https://www.shellcheck.net/wiki/SC2048 - Use "$@" (with quotes) to prevent whitespace problems.
* https://www.shellcheck.net/wiki/SC2124 - Assign as array, or use * instead of @ to concatenate.
* https://www.shellcheck.net/wiki/SC2166 - `-a` is not well defined
* https://www.shellcheck.net/wiki/SC2128 - Expanding an array without an index only gives the first element.
* https://www.shellcheck.net/wiki/SC2086 - Double quote to prevent globbing and word splitting.
* https://www.shellcheck.net/wiki/SC2181 - Check exit code directly with e.g. 'if mycmd;', not indirectly with $?.

### DIFF
--- a/nerves-env.sh
+++ b/nerves-env.sh
@@ -6,41 +6,42 @@
 # "readlink -f" implementation for BSD
 # This code was extracted from the Elixir shell scripts
 readlink_f () {
+    # shellcheck disable=SC2164
     cd "$(dirname "$1")" > /dev/null
     filename="$(basename "$1")"
     if [ -h "$filename" ]; then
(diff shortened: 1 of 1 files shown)


fhunleth:
Interesting. Did you check that it still works? No worries if not. I can check too since I use `nerves-env.sh` all the time.

ConnorRigby (approved):

PR #680, opened 2022-09-22
erlang: bump to 25.1

jjcarstens:
See https://www.erlang.org/patches/otp-25.1

There was a vulnerability referenced in ssl and it was suggested that updating to 25.1 is best.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,1 +1,1 @@
-erlang 25.0.4
+erlang 25.1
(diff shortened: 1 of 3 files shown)


jjcarstens:
Tested on a few devices and this seems to be okay so far

jjcarstens:
Tested this a bit more over the weekend. The potentially breaking changes to `:crypto` seem to be okay in what we typically use.

Also, I tried `libp11 0.4.12` again for kicks and giggles in case there error _might_ have been in OTP when calling and...no dice

fhunleth (approved):

PR #684, opened 2022-10-04
erlang: bump to 25.1.1

jjcarstens:


jjcarstens:
Looks like the `linux/arm64` version of `hexpm/erlang:25.1.1-ubuntu-jammy-20220428` didn't get built for some reason which is why this is failing.

Posted a message to the #hex channel for some help and waiting for a response

jjcarstens:
Looks like the image has been built now! CI is churning and should be ready now

PR #690, opened 2022-11-07
nerves_heart: bump to v2.0.0

fhunleth:


jjcarstens (approved):

jjcarstens:
Maybe we should add a comment to the Config.in stating the need for the Linux expected options?

See [Buildroot manual - 18.2.5](https://buildroot.org/downloads/manual/manual.html#_config_files)

```
CONFIG_WATCHDOG=y
CONFIG_WATCHDOG_NOWAYOUT=y
```

fhunleth:
The Linux kernel options haven't changed from Nerves Heart 1.0. I don't think anyone would see the comment anyway, since the `Config.in` isn't really used. The package can't easily be unselected. See below for a picture. Nerves Heart is like `boardid` and `erlinit` and required:

image

PR #710, opened 2023-02-02
Include Jose's code path caching patch

fhunleth:
This makes it possible to run releases in interactive mode without the
big performance hit from path lookups. Unless a lot of applications
aren't being started, booting in interactive mode should take the same
amount of time as booting in embedded mode.


fhunleth:
Superceded by #712

PR #731, opened 2023-05-16
Bump fwup to 1.10.0 in Dockerfile

fhunleth:


jjcarstens (approved):
Looks like CircleCI is on the struggle 🚌

PR #736, opened 2023-06-27
Don't consume stderr in `merge-squashfs`

jjcarstens:
Part of https://github.com/nerves-project/nerves/issues/877 for adding more useful output during the firmware build process.

https://github.com/nerves-project/nerves/issues/876 shows cases where `mix firmware` fails to build. In reproducing, I found that `merge-squashfs` was failing silently. Specifically at this line because `2>/dev/null` was forwarding stderr to be ignored.

Instead, this changes to let stderr through so more helpful messages can be reviewed

### DIFF
--- a/scripts/merge-squashfs
+++ b/scripts/merge-squashfs
@@ -184,7 +184,7 @@ find_to_pseudofile "$overlay_dir" >>pseudofile.in
 # remove repeated entries to avoid warnings from mksquashfs
 awk '!x[$1]++' pseudofile.in > pseudofile
 
-unsquashfs "$input_squashfs" >/dev/null 2>/dev/null
+unsquashfs "$input_squashfs" >/dev/null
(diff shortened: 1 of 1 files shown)


fhunleth:
What's the error?

My feeling is that redirecting stderr was don't on purpose. I agree that it shouldn't be redirected here, but I think that `merge-squashfs` probably needs to do some stderr redirection now to prevent extraneous output on success cases.

jjcarstens:
In this case - `FATAL ERROR: write_file: file squashfs-root/usr/lib/xtables/libip6t_hl.so already exists`

jjcarstens:
I ran it with a a few different successful builds and it does not seem to output anything (using `unsquashfs 4.6.1`)

fhunleth:
Yes. Thanks!!!

PR #751, opened 2023-09-21
erlang: bump to 26.1

fhunleth:


patches/buildroot/0015-erlang-add-recursive-search-fix.patch
jjcarstens, line 1: I don't think this fix made it into OTP 26.1 and we might still need this patch? (or at least modify it)

fhunleth: I got lazy and combined it with the overall OTP 26.1 patch set. It was a bit of an accident. Separating out the patches is a bit of a pain once they start getting carried between versions for me. It's actually there.

jjcarstens: O, derp. I didn't look hard enough 🤦

jjcarstens (approved):

PR #771, opened 2024-01-18
Support 'find' for MacOS Sonoma

CaptChrisD:
`find` in recent macOS Sonoma -- or possibly due to a Xcode command line tool update -- changed and now matches Linux's.

Instead of leave conditional check from Darwin I am proposing using a simple `||`.

Note: during `mix firmware` if the first `find` fails output to STDERR so a line similar to `find: .../_build/_nerves-tmp/rootfs_overlay/srv/erlang/lib/nautic_net_device-0.1.2/priv/arm/tailscale: unknown primary or operator` will occur.

### DIFF
--- a/scripts/scrub-otp-release.sh
+++ b/scripts/scrub-otp-release.sh
@@ -114,12 +114,9 @@ if [ -n "$NOSCRUBS" ]; then
     # exclusions on MacOS and so if you are attempting to skip scrubbing files,
     # the task will fail. To get the proper variable expanded, we need to
     # eval the call.
-    if [[ $(uname -s) == "Darwin" ]]; then
-        # shellcheck disable=SC2294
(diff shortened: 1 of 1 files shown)


jjcarstens (changes requested):
I think this will need some more work. With my setup, the first `find` passes, but doesn't properly exclude. My system needs the `eval` for whatever reason to properly exclude.

It might be we make this more manual and iterate through executables to remove the ones with a matching `no scrub` directory

fhunleth:
@jjcarstens @CaptChrisD What are your thoughts on this one? I'm ready to make a 1.26.0 release. If you both are too busy, I'm ok with pushing it to a 1.26.1 release.

jjcarstens:
I'm adjusting this right now. It should be a simple fix

fhunleth:
Fixed in #777. Thanks @CaptChrisD for getting this started.

PR #773, opened 2024-01-26
Bump buildroot to 2023.11.1

mlainez:
Hi,

Since the upgrade to v1.25.3 I couldn't run cog inside weston anymore for our in-car infotainment pet project. I tried to fix it using different routes but the one that ended up working was to force nerves_system_br to use the latest available tag for buildroot.

I had to adapt a few things to make nerves work:
- Make modifications in the erlang buildroot package patch
- Remove the pppd patch that used to bump it's version to 2.5.0 since buildroot 2023.11.1 uses this one already
- Create a new patch for libzlib so buildroot uses 1.3.1 since for some reason, libzlib 1.3 was not available for download and returning a 404, breaking the build
- Update the hash for the source tar.gz of libcamera-apps that seemed to missmatch

I'm sending this for reference in case it can help anyone. It's the first time I'm doing PRs on this type of projects so I have no idea if I'm following any of the contribution guidelines. Feel free to suggest improvements if this is useful.

All I know is that nerves still works after this and my issues with weston and cog are fixed.


fhunleth:
This is super helpful!!!

Thank you for splitting up the changes into multiple commits. I'm going to cherry-pick them as I test and integrate them. GitHub might show that there are conflicts while I do this, but just ignore those.

fhunleth:
Your changes were pretty close. I had to make a minor change to the Erlang patch and the new PR is at #774. The other changes I made were:

1. Somehow I downloaded a libcamera-apps.tar.gz with a good checksum so I dropped the patch to fix the checksum. The libcamera-apps.tar.gz is now on the [backup site](https://dl.nerves-project.org/libcamera-apps/libcamera-apps-1.2.1.tar.gz) for comparison. I'll watch this in case there's something weird going on.
2. I put zlib-1.3.tar.xz on the backup site as well. The zlib website always moves their old releases which is a little annoying, but both Buildoot and Nerves have backup sites to work around this. If you were to erase you cached copy and build again, it should work. No need to try this since the official system builds will exercise this a lot.

Thanks again for making the PR. It saves me a lot of time.

mlainez:
Thanks! That was fast.

> 1. Somehow I downloaded a libcamera-apps.tar.gz with a good checksum so I dropped the patch to fix the checksum. The libcamera-apps.tar.gz is now on the [backup site](https://dl.nerves-project.org/libcamera-apps/libcamera-apps-1.2.1.tar.gz) for comparison. I'll watch this in case there's something weird going on.

I was unsure about that one to be honest, it felt weird that the hash had changed but it seemed to me the archive came from the rightful source. I did notice a backup site from nerves in the logs but was unsure at that time if it was just a dumb mirror or something else.

> 2. I put zlib-1.3.tar.xz on the backup site as well. The zlib website always moves their old releases which is a little annoying, but both Buildoot and Nerves have backup sites to work around this. If you were to erase you cached copy and build again, it should work. No need to try this since the official system builds will exercise this a lot.

Indeed, this is annoying, but if buildroot too has a backup sites for this, it seems weird that 1.3 was not on it yet, maybe I was in the unlucky timeframe at that moment.

> 
> Thanks again for making the PR. It saved me a lot of time.

No worries! Have a nice weekend.

PR #778, opened 2024-02-06
Force mktemp to honor $TMPDIR in merge-squashfs script

schrockwell:
This is a proposed solution to https://github.com/nerves-project/nerves/issues/876.

On macOS, `mktemp` will create a temporary directory in `/var/...` which is on a case-insensitive volume. The `merge-squashfs` script will fail when Buildroot packages contain files that differ only in their case.

By adding the `-p ''` argument to `mktemp`, it is forced to honor the `$TMPDIR` environment variable, which the user could specify ahead of time as a special directory on a case-sensitive volume.

### DIFF
--- a/scripts/merge-squashfs
+++ b/scripts/merge-squashfs
@@ -175,7 +175,8 @@ if [[ ! -d $overlay_dir ]]; then
     exit 1
 fi
 
-workdir=$(mktemp -d 2>/dev/null || mktemp -d -t 'merge-squashfs-tmp')
+# force mktemp to honor $TMPDIR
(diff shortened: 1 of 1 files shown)


fhunleth:
Adding `-p ''` seems reasonable regardless. It took me a bit to see if this change has any portability issues, but I really think it's safe.

Regarding using `$TMPDIR` as a workaround for the case-insensitivity issue, I think this is the most practical near-term solution. Getting rid of the need for `merge-squashfs` is my favorite one, but that's definitely more work.

Thank you for the PR.

Issue #781, opened 2024-02-25
Firmware builds broken on macOS 13


th0mas:
### Environment

* Elixir version (`elixir -v`): `Elixir 1.15.4 (compiled with Erlang/OTP 26)`
* Nerves environment: 
```
Nerves environment
MIX_TARGET:   rpi3
MIX_ENV:      dev
```
* Additional information about your host, target hardware or environment that
  may help
  * macOS 13.0, running fish shell

### Current behavior

Running `mix firmware` errors with
```
Updating base firmware image with Erlang release...
Copying rootfs_overlay: /Users/tom/dome_radio/dome_radio_firmware/_build/rpi3_dev/nerves/rootfs_overlay
Copying rootfs_overlay: /Users/tom/dome_radio/dome_radio_firmware/rootfs_overlay
mktemp: illegal option -- p
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix
```

### Expected behaviour
* Firmware builds correctly.

It looks like e6135acab77375853b9672462a19eb7c81d870b4 introduced an unsupported `mktemp` flag, breaking builds on macOS

fhunleth:
That's odd since that commit was specifically for MacOS users. Could you check where mktemp comes from? I wonder if it's built into a shell and for whatever reason you have an old version?

We could certainly add a fallback for your case since I'm sure you're not alone. Given the history of this line, I'm curious to understand better what's going on.

th0mas:
Hmmm - it looks like macOS 14 supports the `-p` flag, it's only versions 13 and below that don't seem to support the `-p` flag.

I'll use this as an excuse to finally update my laptop 😅

guillego:
I'm seeing the same error message in macOS 13 when I run mix firmware after bumping my `nerves_system_br` version. I'll also take it as a cue to upgrade to macOS 14 😅.

<details><summary>
This is the manpage for the current mktemp</summary>
<p>
MKTEMP(1)                   General Commands Manual                  MKTEMP(1)

NAME
     mktemp – make temporary file name (unique)

SYNOPSIS
     mktemp [-d] [-q] [-t prefix] [-u] template ...
     mktemp [-d] [-q] [-u] -t prefix

DESCRIPTION
     The mktemp utility takes each of the given file name templates and
     overwrites a portion of it to create a file name.  This file name is
     unique and suitable for use by the application.  The template may be any
[...]

PR #782, opened 2024-03-04
Add explicit check for $TMPDIR before calling mktemp in merge-squashfs

schrockwell:
This is a follow-up to #778 and a proposed fix for #781.

Here's what I know:

- `mktemp` on macOS 13 does not support the `-p` option (as reported in #781)
- The `-t` option is problematic
  - The option behaves wildly differently between macOS and Linux
  - Ubuntu 18 manpages claim the `-t` option is deprecated
  - The original fallback invocation of `mktemp -d -t 'merge-squashfs-tmp'` doesn't even work on Linux; the template string requires "X" characters as placeholders
- `mktemp` does not consistently honor `$TMPDIR`
  - Linux `mktemp` honors `$TMPDIR` by default
  - macOS `mktemp` does NOT honor `$TMPDIR` by default

So, my proposed solution avoids `-t` and `-p` and is about as simple and explicit as possible, and it has been tested as working equally well on macOS 13, 14, and Ubuntu 18, and with or without `$TMPDIR` being set.

### DIFF
--- a/scripts/merge-squashfs
+++ b/scripts/merge-squashfs
@@ -175,8 +175,13 @@ if [[ ! -d $overlay_dir ]]; then
     exit 1
 fi
 
-# force mktemp to honor $TMPDIR
-workdir=$(mktemp -d -p '' 2>/dev/null || mktemp -d -p '' -t 'merge-squashfs-tmp')
+# Use $TMPDIR if it's set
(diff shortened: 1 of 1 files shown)


jjcarstens (approved):

fhunleth:
Thanks again for fixing this!

PR #791, opened 2024-05-12
Manually create skel

ringlej:
With https://github.com/nerves-project/nerves/pull/972, `useradd` fails to copy the /etc/skel to the new home directory.
This fixes by manually doing this copy.

### DIFF
--- a/create-build.sh
+++ b/create-build.sh
@@ -67,6 +67,16 @@ if [[ $HOST_ARCH != "x86_64" ]] && [[ $HOST_ARCH != "aarch64" ]]; then
     exit 1
 fi
 
+# Check that NERVES_BUILD_DIR is on a case-sensitive filesystem
+touch case.sensitive
+touch CASE.sensitive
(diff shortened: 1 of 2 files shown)


fhunleth:
What error does `useradd` give? This seems really strange. I'm not opposed to merging, but I'm curious as to what happened.

ringlej:
Whe the https://github.com/nerves-project/nerves/pull/972 PR, the `/home/nerves/` directory gets created for the volume mount `/home/nerves/.ssh`. This happens before the `docker-entrypoint.sh` script is executed. When `useradd` is run with the `-m` option, it is asking `useradd` to copy the home directory template at `/etc/skel/`, but it will only do so if the new user's home directory does not exist. It produces this warning:

```
useradd: warning: the home directory /home/nerves already exists.
useradd: Not copying any file from skel directory into it.
```
Changing the `useradd` option `-m` to a `-M` tells it to not bother copying the template at `/etc/skel/` to the new user's home directory, and we go ahead and do this with `cp -pr /etc/skel/. ~nerves`

fhunleth:
Oh, I understand now. That makes sense. Thank you for explaining it to me.

PR #794, opened 2024-05-20
erlang: bump to 27.0

fhunleth:


fhunleth:
I tested on rpi4, rpi0 and bbb. Seems to work well so far.

PR #799, opened 2024-06-23
buildroot: bump to 2024.05

fhunleth:
This bumps Buildroot from 2024.02.2 to 2024.05.

From the [2024.05 announcement](https://lore.kernel.org/buildroot/87bk46tjk2.fsf@dell.be.48ers.dk/T/):

```
Of noteworthy new features/changes we have:

Toolchains:
- Binutils 2.42 added, 2.39 dropped, 2.41 is now default

- GCC 14.x added, 11.x dropped, 13.x is now default

- uClibc-ng bumped to 1.0.48

- Musl bumped to 1.2.5, adds riscv32 support

- Glibc bumped to 2.39, libcrypt no longer available (see libxcrypt
  package)

- External glibc based toolchains no longer install libcrypt files
  (except for libcrypt.so.1) to not conflict with the libxcrypt package

- Bootlin external toolchains updated to 2024.02

Architecture:
- Support for riscv32 noMMU added

- Internal toolchains for SPARC (re-)added

- Support for the Motorola m68k specific shared library support for the
  FLAT binary format dropped, as support has been removed from the Linux
  kernel as of v5.19.

Download (notice, Buildroot generated tarballs have changed):
- Tar 1.35 is now required (will be automatically built if not available
  on the host). Notice that tar 1.35 changes the way devmajor/minor is
  stored, so tarballs created with 1.35 are NOT binary identical to
  <1.35

- A custom tar program specified with make TAR=path/to/custom/tar is now
  correctly used everywhere.

- The tarball creation logic has now been extended to correctly handle
  when ACLs are set on the download directory.

- An issue with conflicting git tags between (different versions of) a
  package has been fixes

- Git attributes are now handled correctly

- Fixed an reproducibility issue with the generated archives (for
  git/svn/cargo/go) when multiple packages share an archive (E.G. linux
  vs linux-headers).

- The (noop) --passive-ftp has been removed from the default wget
  options and packages using a FTP upstream have been migrated to
  HTTP(S) for compatiblity with distributions that have moved to wget2
  (which does not support FTP or the --passive-ftp option).

Security:
- The /dev/shm entry in the default /etc/fstab used for busybox and sysv
  init systems was missing the sticky bit, allowing users to delete and
  replace other users files.

New defconfigs: Beagleboneai64, Orangepi Zero2w, Pine64 star64

New packages: basu, bats-assert, bats-file, bats-support, bcc, bpftrace,
export-subst, flutter-adaptive-scaffold-example,
flutter-animations-example, flutter-dynamic-layouts-example,
flutter-go-router-example, flutter-image-example,
flutter-markdown-example, flutter-packages, flutter-rfw-local-example,
go-bootstrap-stage3, igt-gpu-tools, libgpiod2, libxmlb, libzenoh-c,
lmdb, lua-lightningmdb, lua-lsdbus, mokutil, python-email-validator,
python-expandvars, python-fastui, python-hid, python-jc, python-mpmath,
python-poetry-core, python-sympy, python-yamllint, qbee-agent,
starfive-spltool, ustreamer

Removed packages: at91bootstrap, at91dataflashboot, flutter-gallery,
omxplayer, on2-8170-libs, on2-8170-modules, ti-k3-image-gen
```

Most Nerves systems will need to set the version of GCC in the
`nerves_defconfig` like this now:

```
BR2_TOOLCHAIN_EXTERNAL_GCC_13=y
```


fhunleth:
I tested this with a Nerves web kiosk using WPEWebkit on a RPi4 by playing back YouTube videos. That seems pretty good.

jjcarstens (approved):
LGTM. Though I suspect we would have to apply the a similar change for building the release image as well. 


The release notes seem to indicate tar 1.35 is only built if it doesn't exist on host - we could preemptively build it in the images under a user rather than disabling protections and requiring Buildroot to make it?

fhunleth:
I expect that if it's not `tar` that it will be another GNU tool soon that will reject being built as root. I would assume we're getting warnings from other things, but I haven't combed the logs.

I'll leave this here and not merge. I hope there's a way to make non-root builds work since that just seems better all around.

PR #807, opened 2024-07-30
Run unsquashfs with force flag when merging rootfs

bjyoungblood:
This is a temporary fix for an issue that prevents building firmware on
macOS if the Nerves system contains any files with names that collide on
a case-insensitive filesystem like APFS. See nerves-project/nerves#876.

When running `mix firmware`, warnings in the following format may
indicate that some things may not work as expected:

```
Pseudo modify file "path/to/file" does not exist in source filesystem.  Ignoring.
```


fhunleth:
Thanks!

PR #825, opened 2024-11-17
Add TPM support

amclain:
This PR contains a few modifications for a [Trusted Platform Module](https://en.wikipedia.org/wiki/Trusted_Platform_Module) (TPM) to work with Nerves:

- The `tpm2-tss-engine` package was added so that the BEAM can communicate with the TPM via an OpenSSL engine.
- `tpm2-tss` is a core library for communicating with TPMs. v3 was incompatible, so we patched it to v4.
- We pulled in a patch to support compilation with GCC 14.

The Elixir TPM library is available here:
https://github.com/redwirelabs/tpm

This is also compatible with the open source [tpm2-tools](https://github.com/tpm2-software/tpm2-tools) tooling.

### DIFF
--- a/Config.in
+++ b/Config.in
@@ -38,6 +38,7 @@ source "$BR2_EXTERNAL_NERVES_PATH/package/rpicam-apps/Config.in"
 source "$BR2_EXTERNAL_NERVES_PATH/package/rpi-distro-bluez-firmware/Config.in"
 source "$BR2_EXTERNAL_NERVES_PATH/package/rpi-distro-firmware-nonfree/Config.in"
 source "$BR2_EXTERNAL_NERVES_PATH/package/rpi-libcamera/Config.in"
+source "$BR2_EXTERNAL_NERVES_PATH/package/tpm2-tss-engine/Config.in"
 
 # Load user options
 menu "User-provided options"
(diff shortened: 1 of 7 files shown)


fhunleth:
@amclain Have you submitted these upstream to Buildroot? I don't mind temporarily maintaining them here while the Buildroot devs review/merge them. I just want to track the upstream PR so I know when to remove these packages.

amclain:
@abelino is working through [submitting a patch](https://buildroot.org/downloads/manual/manual.html#submitting-patches) to Buildroot, since they do this on the mailing list rather than in a GitHub PR.

Normally we would be happy to maintain this code ourselves, but using a fork of `nerves_system_br` prevents us from being able to publish the Nerves systems that use it to Hex.

fhunleth:
Great, thanks for submitting a patch upstream to Buildroot. Yes, I think I mentioned this before, but I view it as a feature that patches are upstreamed to Buildroot since they can do far better review and testing than I can.

abelino:
I have started the up-streaming process. The patch series can be found here: https://patchwork.ozlabs.org/project/buildroot/list/?series=434002

fhunleth:
@abelino Thanks!!! 

@amclain Could you rebase so that this is two commits, one for `tpm2-tss-engine` and the other for the patch? It all looks good so I'll merge once I see the update.

amclain:
@fhunleth updated. Thanks!

fhunleth:
Thanks!

PR #832, opened 2025-01-08
Add option to rel2fw.sh to allow post-processing

lawik:
This would typically be invoked by `mix firmware` further up the stack. It is at the right moment to allow operating on the squashfs file system before it is packed up into the fw file.

The primary purpose is to allow signing the root filesystem for use with dm-verity.


fhunleth:
Looks good. I like that it's a simple hook.

PR #853, opened 2025-06-16
erlang: bump to 27.3.4.1

fhunleth:


fhunleth:
We're ahead of hexbob in building. Note to self: Check back at https://hub.docker.com/r/hexpm/erlang/tags?name=27.3.4.1- to see when to retry build.

PR #922, opened 2026-04-22
Activate imx500 camera to enable AI for rpicam tool

nomtrosk:
We should also enable opencv but that seem to be a big one to ship with buildroot system


lawik:
You should be getting opencv from evision typically. But not sure if that interops well with this camera..

nomtrosk:
I think that OpenCV should be installed with buildroot in this system if we enable `-Denable_opencv=enabled`, which I think is out of scope for this base system? But I think it's fair to enable support for the IMX500 AI camera. What do you think?

nomtrosk:
Any good pointers why the CI fails? The logs do not show much?

PR #948, opened 2026-08-22
Nerves systems can provide their own buildroot patches

bjyoungblood:
This allows a Nerves system to provide additional Buildroot patches on
top of the ones in this repository via `<system_repo>/patches/buildroot`.

Sometimes, you need a bleeding-edge version of a package or to change
some knobs Buildroot doesn't expose. And sometimes, vendoring that
package is a footgun. For example, if you vendor WPEWebkit and compile
it with multimedia support via gstreamer, `savedefconfig` adds
`# BR2_PACKAGE_GST1_PLUGINS_BAD_PLUGIN_WPE is not set` to your
defconfig, and you can't add it back because it depends on
`BR2_PACKAGE_WPEWEBKIT`.


scripts/buildroot-state.sh
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 # SPDX-FileCopyrightText: 2016 Frank Hunleth
 # SPDX-FileCopyrightText: 2019 Matt Ludwigs
+# SPDX-FileCopyrightText: 2026 Matt Ludwigs
```
fhunleth: Did you get this from Matt or does your Claude think that you're Matt?

bjyoungblood: No Claude on this one. Yesterday was just... a day 😅

lawik: Humans making mistakes? Absurd! <3

fhunleth:
Summarizing off-GitHub discussions: PRs like this have been rejected several times in the past to minimize propagation of non-upstreamed patches and because alternative package recipes aren't that hard to write. This case is different. First, making an alternative package is actually hard for this one. Second, we've been discussing decoupling nerves_system_br from fixed Buildroot versions to allow more flexibility in which Nerves systems get used in a project. Assuming we do the latter, pulling in a custom Buildroot tree "should" be easy and there will be a route to customizing Buildroot as much as anyone desires. That makes this patch a short term workaround until the better way is available.

PR #960, opened 2026-09-23
Stop leaking toolchain host libraries into LD_LIBRARY_PATH

CJRChang:
Fakeroot's default behaviour of prepending toolchain lib dir shadowed host libraries with unversioned Buildroot copies (e.g. ld.so warning "no version information available" on Arch for /bin/sh -> libreadline -> libncursesw) resulting in bad builds.

This change patches fakeroot to preload libfakeroot.so via a Buildroot package patch so builds are (hopefully) the same across platforms:
- removes the prepends
- patch buildroot's fakeroot to use `USEABSLIBPATH=1` to preload `libfakeroot.so` by absolute path

The rationale is an issue on Arch where, when building downstream firmware, I was seeing warnings like:
```
/usr/bin/sh: /home/rc/.nerves/artifacts/nerves_system_rpi4-portable-2.1.2/host/lib/libncursesw.so.6: no version information available (required by /usr/lib/libreadline.so.8)
```
resulting in bad firmware builds.
I had been naively working around this by copying the host system's libncursesw.so.6 into the built artefacts - this "worked" because the host system's libncursesw.so.6 had version definitions but was a terrible hack.

On investigation:
- nerves_system_br prepends the toolchain/system lib directory to LD_LIBRARY_PATH (git history suggests this might be from very early days when using crosstool-ng?)
- upstream fakeroot has `USEABSLIBPATH=1` to load by absolute path but doesn't enable it for linux
 which results in the shadowing and the warning above.

docker (built on debian) builds _don't_ seem to hit this because on debian:
- ncurses is split into libtinfo6 and libncursesw6
- libtinfo6 isn't shipped by buildroot so no shadowing occurs and it uses the container's libtinfo6
- readline happily links with libtinfo6 so no warning occurs

I've confirmed this change results in a clean build on Arch but I haven't had a chance to test on other platforms.


fhunleth:
@CJRChang Thanks for sending this! I expect to get back to merging PRs after the Goatmire conference.

PR #961, opened 2026-09-23
Speed up merge-squashfs pseudofile generation

bjyoungblood:
unsquash_to_pseudofile and find_to_pseudofile shelled out to sed for
every line, while find ran stat for every file. These invokations have a
much more pronounced impact on macOS than Linux due to macOS's slow
`fork()`.

This replaces the sed calls with bash expansion and makes find run stat
in batches.

Tested with bash 3.2 (macOS) and 5.2 (Linux) with a fixture containing a
mixture of regular files, device nodes, setuid/setgid/sticky bits, names
containing spaces, and similar potential hazards.

Against a real Nerves project with a ~214 MB rootfs, this cuts merge
time on macOS by more than half, and Linux sees a decent improvement as
well.

  macOS:        24.8s -> 9.3s (-63%)
  Linux arm64:  11.6s -> 8.2s (-29%)

Assisted-by: Claude Code:claude-opus-5.5


fhunleth:
Hi @bjyoungblood. Thanks for sending this and especially for taking the time to write the PR message. 

Nerves 2.0 should be even faster as it removes this code entirely. I'm feeling like it would be better to start using the Nerves 2.0.0-pre.2 branch instead. This is annoying tricky code, and it's not unit tested sufficiently to give me a warm fuzzy (my fault, not yours). You have a faster mac than I do, but to give you an idea on Nerves 2, the Nerves web kiosk demo went from about 26s to 8.5s for incremental builds for me. Plus Nerves 2 has edge case tests for the things I've run into, so I feel more comfortable that I can accept changes there.

See https://github.com/nerves-web-kiosk/kiosk_demo/commit/4af3148a098b11d4a87f278a347cca2a689c1069 for how to update. Your Nerves system needs to output a `.tar` file to avoid needing to install sqfs2tar. Even though it's a `-pre` release, your use of it should be pretty low risk for future changes.

bjyoungblood:
This is all I have to say about that:

img

